PR #19, opened 2021-03-28
Add support for Python 3.9

berendkleinhaneveld:
Dicts have gained `__or__`, `__ror__` and `__ior__` functions.
Those functions are now wrapped as well.

There is also the new `__class_getitem__` function which is now
part of the excluded functions.

Bonus: you can now pass some arguments to `pytest` through the poetry `scripts.py`.


Korijn:
Thanks! <3

PR #20, opened 2021-03-28
Add qt example

berendkleinhaneveld:
I took the liberty of constructing an example that shows how `observ` can be combined with PyQt. Hope you like it.

One thing that I noticed is that passing `immediate=True` to the `watch` function didn't trigger the watcher, because the watcher is not 'dirty'. And that is because the `dirty` flag is set to the value of `lazy`, which defaults to `False` in the `watch` function. 
Hence I added the line in the `watch` function to ensure that the watcher is evaluated when passing `immediate=True`. Small problem is that `old_value` and `new_value` in the callback will be the same value on first evaluation...

I've also added a line with PyQt in pyproject.toml to easily run the Qt example, but I've left it commented out, because I don't think it is really a 'dev' dependency.

### DIFF
--- a/examples/observe_qt.py
+++ b/examples/observe_qt.py
@@ -31,7 +31,7 @@ class Display(QWidget):
                 return "Please click the button below"
             return f"Clicked {state['clicked']} times!"
 
-        self.watcher = watch(lambda: label_text(), self.update_label, immediate=True)
+        self.watcher = watch(label_text, self.update_label, immediate=True)
 
     def update_label(self, old_value, new_value):
         self.label.setText(new_value)
--- a/examples/observe_qt.py
+++ b/examples/observe_qt.py
@@ -1,0 +1,82 @@
+"""
+PyQt example that shows how to utilize observ
+in combination with Qt.
+
+State is passed to the widgets. One of the widgets
+adjusts the state while the other watches the state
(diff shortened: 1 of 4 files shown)


examples/observe_qt.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.setLayout(layout)
+
+        @computed
+        def label_text():
+            if state["clicked"] == 0:
+                return "Please click the button below"
+            return f"Clicked {state['clicked']} times!"
+
+        self.watcher = watch(lambda: label_text(), self.update_label, immediate=True)
```
Korijn: I guess the API could be simplified a bit so you can write `watch(label_text, ...` 🤔

Korijn:
Guess the test suite needs some love too now.

Awesome example though, thanks a lot! It's a great demonstration of observ. 🤗

examples/observe_qt.py
Korijn: I didn't realize it already supported this syntax! Awesome! <3

Korijn:
LGTM! Let's merge!

Korijn:
@berendkleinhaneveld Just a thought here... I think you don't need the `@computed` decorator here. It should already work just by watching the expression. I didn't consider that before. 🤔

PR #26, opened 2021-09-21
Scheduler

### LINKED ISSUE #18: Callback scheduling
Should they be buffered? Should I provide event loop integration for various libs (i.e. asyncio, qt, etc)?

Korijn:
Closes #18, #25

**To do**

- [x] Switch example to PySide6
- [x] Split library into modules
- [x] Sprinkle docstrings
- [x] Fix test suite


pyproject.toml
```diff
@@ -18,7 +18,7 @@ flake8-import-order = "^0.18.1"
 flake8-print = "^3.1.4"
 pytest = "^5.4.3"
 pytest-cov = "^2.10.0"
-# PyQt5 = "^5.15.4"  # needed for PyQt example
+PySide6 = "^6.0.2"
```
berendkleinhaneveld: @Korijn Do you want to keep this commented out? I guess this was commented out to save some time on CI?

Korijn: Nope, this is fine! Not sure why it was commented out

pyproject.toml
```diff
@@ -8,7 +8,7 @@ homepage = "https://github.com/Korijn/observ"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.6"
+python = ">=3.6,<3.10"
```
Korijn, line 11: At first I didn't like this, but technically it is true, we don't test for python 3.10 and our observable collections definitely need it!

observ/__init__.py
```diff
@@ -47,31 +47,102 @@ def _traverse(obj, seen):
...
+
+        self.has.add(watcher.id)
+        if not self.flushing:
+            self._queue.append(watcher)
+        else:
+            # If already flushing, splice the watcher based on its id
+            # If already past its id, it will be run next immediately.
+            i = len(self._queue) - 1
+            while i > self.index and self._queue[i].id > watcher.id:
+                i -= 1
```
Korijn: If we also track a list of watcher ids while queueing we can cheaply use bisect.insort here

examples/observe_qt.py
```diff
@@ -56,18 +98,43 @@ def __init__(self, state, *args, **kwargs):
...
-    state = observe({"clicked": 0})
+    state = observe({"clicked": 0, "progress": 0})
 
     app = QApplication([])
 
+    # Use a timer to run the scheduler
+    timer = QTimer()
+    timer.timeout.connect(scheduler.flush)
+    timer.setInterval(1000 / 60)
+    timer.start()
```
Korijn: In order to avoid event queue build up (when flush takes longer to run than 1000/60) we should instead schedule a flush only when necessary (on queueing), and not reschedule while it is still pending

Korijn:
~~Fix this (happens after closing the app):~~ Fixed by removing teardown!

image

PR #29, opened 2021-09-24
Scheduler unit tests

Korijn:
Had to refactor the scheduler slightly to make it easier to test:

- Make it possible to configure noop request flush handler, so you can call flush manually in tests
- Make it possible to clear the scheduler manually with `scheduler.clear()` so you can reset state after tests (necessary since the scheduler is global)

Added tests for the following cases:

- Assert exception is raised when no request flush handler is registered
- Assert exception is raised when cyclical dependency in watched expression
- Assert exception is raised when cyclical dependency via watcher callback
- Assert exception is raised when cyclical dependency exists indirectly between two watchers
- Assert flush works in basic setting
- Assert flush works when a callback triggers another watcher (queue growth)

Also adds:

- `scheduler.detect_cycles` just a toggle for cycle detection, so you can disable it in prod

### DIFF
--- a/observ/__init__.py
+++ b/observ/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.4.1"
+__version__ = "0.4.2"
 
 
 from .api import *
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -15,6 +15,19 @@ class Scheduler:
         self.circular = {}
         self.index = 0
         self.waiting = False
+        self.request_flush = self.request_flush_raise
+
+    def request_flush_raise(self):
+        """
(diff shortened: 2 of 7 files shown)


Korijn:
I decided to add one more test and found a hole in our infinite loop detection. See `test_queue_growth_branching`.

I was intending to create a test that checks that when you have if-statements in your watched expression, that can trigger more watchers to go into the queue _while_ you are flushing, and that all resolves properly.

Instead I set up the test wrong, and then triggered an infinite loop. Guess we missed a case, that might also be a problem in Vue?!

Korijn:
Fixed!

PR #30, opened 2021-09-24
Support 0-2 arguments for callbacks

berendkleinhaneveld:
Here's my attempt at supporting callbacks with 0-2 arguments.

Instead of having to add (and name) an extra argument in the `watch` function, I decided to try a different route (which also supports callbacks without arguments).
We also played with the idea of specifying how many arguments the callback has, but that has the downside of being verbose in some cases.

What if the watcher just tries out which callback works and then take note of how many arguments it used? Seems to work and doesn't give any overhead for repeated calls to `run_callback` (I did ran some benchmarks to verify this).

Before running some benchmarks, I had some 'smart' code for passing the right amount of arguments to the callback (`self.callback(*[old, new][2 - nr :])`), but the benchmarks pointed out that a simple `if ... == ...` test was significantly faster. 

So, what do you think?

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -4,6 +4,7 @@ observable datastructures, and optionally trigger callback when
 a change is detected.
 """
 from collections.abc import Container
+import inspect
 from itertools import count
 from typing import Any
 from weakref import WeakSet
@@ -33,6 +34,15 @@ def _traverse(obj, seen):
 # keep track of the order in which subscribers will
 # be notified
 _ids = count()
+
+
+class WrongNumberOfArgumentsError(TypeError):
+    """
+    Error that is used to signal that the wrong number of arguments is
+    used for the callback
+    """
+
+    pass
 
 
 class Watcher:
@@ -74,9 +84,9 @@ class Watcher:
             old_value = self.value
             self.value = value
             if self.callback:
-                self.run_callback(old_value, self.value)
+                self.run_callback(self.value, old_value)
 
-    def run_callback(self, old, new):
+    def run_callback(self, new, old):
         """
         Runs the callback. When the number of arguments is still unknown
         for the callback, it will fall into the try/except contstruct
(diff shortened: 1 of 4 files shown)


observ/watcher.py
```diff
@@ -73,7 +74,37 @@ def run(self) -> None:
...
+        to figure out the right number of arguments.
+        After running the callback one time, the number of arguments
+        is known and the callback can be called with the correct
+        amount of arguments.
+        """
+        if self._number_of_callback_args == 1:
+            self.callback(new)
+            return
+        if self._number_of_callback_args == 2:
+            self.callback(old, new)
```
Korijn: I feel like maybe new, old, makes more sense. Do we swap or leave it as it is?

Korijn:
I guess the only "gotcha" we introduce here is if a user callback raises a typeerror, that that will be masked by our handler (during the first run). I don't know how to deal with that cleanly (by which I mean I do know dirty tricks with stack inspection but... No).

Awesome job though, I think we should merge it!

observ/watcher.py
berendkleinhaneveld: Yeah, let's swap it, had some doubts on whether it was worth the effort but now would be a good time.

berendkleinhaneveld:
> I guess the only "gotcha" we introduce here is if a user callback raises a typeerror, that that will be masked by our handler (during the first run). I don't know how to deal with that cleanly (by which I mean I do know dirty tricks with stack inspection but... No).

We could check the exception to see if it is about the number of arguments... 🤔

Korijn:
> > I guess the only "gotcha" we introduce here is if a user callback raises a typeerror, that that will be masked by our handler (during the first run). I don't know how to deal with that cleanly (by which I mean I do know dirty tricks with stack inspection but... No).
> 
> We could check the exception to see if it is about the number of arguments... 🤔

Thing is, the user callback could also throw exactly the same TypeError, message and everything. This is the achilles heel of exception handling, really. :) I'm not sure if we can perfectly handle this situation without resorting to stack trace inspection (to check if the exception was raised on the same stack depth). That would be using `inspect.trace()` probably.

I'm not sure how expensive this is, but we cache the result anyway!

Here's an example on how to do it:

```python
import inspect


def foo(a, b):
    print(a + b)


def bar(a, b):
    foo(a)  # typeerror


def baz(a, b):
    foo(a, b)


def make_call(callback, *args):
    try:
        callback(*args)
    except TypeError:
        frames = inspect.trace()
        try:
            # from https://docs.python.org/3/library/inspect.html#inspect.trace
            # The first entry in the list represents the caller; the last entry represents where the exception was raised.
            # so: length 1 means the caller frame and raised frame are the same
            # and therefore it must have been our call that raised the type error
            # otherwise it has a different origin and the exception should be reraised
            if len(frames) != 1:
                raise
        finally:
            del frames  # avoid reference cycles, see Note at https://docs.python.org/3/library/inspect.html#the-interpreter-stack


def main():
    make_call(baz, 5)  # local typeerror
    print("handled local type error")

    try:
        make_call(bar, 5, 6)  # user callback typeerror
    except TypeError:
        print("user type error re-raised")

    make_call(baz, 5, 6)  # no typeerror
    print("good call")


if __name__ == "__main__":
    main()
```

Output:

```
handled local type error
user type error re-raised
11
good call
```

berendkleinhaneveld:
Aha, now I gotcha!

> I'm not sure how expensive this is, but we cache the result anyway!

I'll check!

berendkleinhaneveld:
I think I actually would like to hide the WrongNumberOfArgumentsError to the user by just raising the original TypeError instead. As that should contain all the info they need to know.

Korijn:
> I think I actually would like to hide the WrongNumberOfArgumentsError to the user by just raising the original TypeError instead. As that should contain all the info they need to know.

That makes sense! Open another PR? :)

PR #32, opened 2021-09-27
Proxies

### LINKED ISSUE #16: Minimize deps
E.g. dict.copy causes depend to be called for the dict dep and all its keydeps, where only the dict dep would be sufficient.

This mostly matters in eager mode but in general it would shave off some redundant function calls

### LINKED ISSUE #23: Proxying
It should be (relatively) easy to also add `readonly(state)` which would provide a proxy onto an observable that throws an error on any write operation.

Korijn:
Closes #23
Closes #16

This PR changes the reactivity mechanism so that you can have multiple proxy references to a single object, and such that objects are proxied lazily. This is based on Vue 3's approach, and allows for readonly proxies.

- [x] toRaw equivalent
- [x] verify readonly proxies
- [x] use readonly proxies for computed
- [x] documentation of caveats (i.e. gc)
- [x] update README

### DIFF
--- a/observ/api.py
+++ b/observ/api.py
@@ -4,12 +4,12 @@ Defines the public API for observ users
 from functools import wraps
 
 from .dep import Dep
-from .observables import observe
+from .observables import reactive
 from .scheduler import scheduler
 from .watcher import Watcher
 
 
-__all__ = ("observe", "computed", "watch", "scheduler")
+__all__ = ("reactive", "computed", "watch", "scheduler")
 
 
 def computed(fn):
--- a/observ/api.py
+++ b/observ/api.py
@@ -13,7 +13,7 @@ __all__ = ("reactive", "computed", "watch", "scheduler", "to_raw")
 
 
 def computed(fn):
-    watcher = Watcher(fn)
+    watcher = Watcher(fn, readonly=True)
 
     @wraps(fn)
     def getter():
--- a/observ/api.py
+++ b/observ/api.py
@@ -13,7 +13,7 @@ __all__ = ("reactive", "computed", "watch", "scheduler", "to_raw")
 
 
 def computed(fn):
-    watcher = Watcher(fn, readonly=True)
+    watcher = Watcher(fn)
 
     @wraps(fn)
     def getter():
--- a/observ/api.py
+++ b/observ/api.py
@@ -1,15 +1,24 @@
 """
 Defines the public API for observ users
 """
-from functools import wraps
+from functools import partial, wraps
 
 from .dep import Dep
-from .observables import reactive, to_raw
+from .observables import proxy, to_raw
 from .scheduler import scheduler
 from .watcher import Watcher
 
 
-__all__ = ("reactive", "computed", "watch", "scheduler", "to_raw")
+__all__ = (
+    "reactive",
+    "readonly",
+    "shallow_reactive",
+    "shallow_readonly",
+    "computed",
+    "watch",
+    "scheduler",
+    "to_raw",
+)
 
 
 def computed(fn):
@@ -35,3 +44,9 @@ def watch(fn, callback, sync=False, deep=False, immediate=False):
         if watcher.callback:
             watcher.run_callback(watcher.value, None)
     return watcher
+
+
+reactive = partial(proxy)
+readonly = partial(proxy, readonly=True)
+shallow_reactive = partial(proxy, shallow=True)
+shallow_readonly = partial(proxy, shallow=True, readonly=True)
--- a/observ/api.py
+++ b/observ/api.py
@@ -4,12 +4,12 @@ Defines the public API for observ users
 from functools import wraps
 
 from .dep import Dep
-from .observables import reactive
+from .observables import reactive, to_raw
 from .scheduler import scheduler
 from .watcher import Watcher
 
 
-__all__ = ("reactive", "computed", "watch", "scheduler")
+__all__ = ("reactive", "computed", "watch", "scheduler", "to_raw")
 
 
 def computed(fn):
--- a/observ/api.py
+++ b/observ/api.py
@@ -46,7 +46,7 @@ def watch(fn, callback, sync=False, deep=False, immediate=False):
     return watcher
 
 
-reactive = partial(proxy)
+reactive = proxy
 readonly = partial(proxy, readonly=True)
 shallow_reactive = partial(proxy, shallow=True)
 shallow_readonly = partial(proxy, shallow=True, readonly=True)
--- a/observ/observables.py
+++ b/observ/observables.py
@@ -182,6 +182,14 @@ shallow_reactive = partial(proxy, shallow=True)
 shallow_readonly = partial(proxy, shallow=True, readonly=True)
 
 
+class StateModifiedError(Exception):
+    """
+    Raised when a proxy is modified in a watched (or computed) expression.
+    """
+
+    pass
+
+
 def read_trap(method, obj_cls):
     fn = getattr(obj_cls, method)
 
@@ -240,7 +248,7 @@ def write_trap(method, obj_cls):
     @wraps(fn)
     def inner(self, *args, **kwargs):
         if Dep.stack:
-            raise ReadonlyError()
+            raise StateModifiedError()
         args = tuple(proxy(a) for a in args)
         kwargs = {k: proxy(v) for k, v in kwargs.items()}
         retval = fn(self.target, *args, **kwargs)
@@ -258,7 +266,7 @@ def write_key_trap(method, obj_cls):
     @wraps(fn)
     def inner(self, *args, **kwargs):
         if Dep.stack:
-            raise ReadonlyError()
+            raise StateModifiedError()
         key = args[0]
         is_new = key not in proxy_db.attrs(self)["keydep"]
         old_value = getitem_fn(self.target, key) if not is_new else None
@@ -282,7 +290,7 @@ def delete_trap(method, obj_cls):
     @wraps(fn)
     def inner(self, *args, **kwargs):
         if Dep.stack:
-            raise ReadonlyError()
+            raise StateModifiedError()
         retval = fn(self.target, *args, **kwargs)
         proxy_db.attrs(self)["dep"].notify()
         for key in self._orphaned_keydeps():
@@ -299,7 +307,7 @@ def delete_key_trap(method, obj_cls):
     @wraps(fn)
     def inner(self, *args, **kwargs):
         if Dep.stack:
-            raise ReadonlyError()
+            raise StateModifiedError()
         retval = fn(self.target, *args, **kwargs)
         key = args[0]
         proxy_db.attrs(self)["dep"].notify()
@@ -322,6 +330,10 @@ trap_map = {
 
 
 class ReadonlyError(Exception):
+    """
+    Raised when a readonly proxy is modified.
+    """
+
     pass
 
 
--- a/observ/observables.py
+++ b/observ/observables.py
@@ -239,6 +239,8 @@ def write_trap(method, obj_cls):
 
     @wraps(fn)
     def inner(self, *args, **kwargs):
+        if Dep.stack and Dep.stack[-1].readonly:
+            raise ReadonlyError()
         args = tuple(proxy(a) for a in args)
         kwargs = {k: proxy(v) for k, v in kwargs.items()}
         retval = fn(self.target, *args, **kwargs)
@@ -255,6 +257,8 @@ def write_key_trap(method, obj_cls):
 
     @wraps(fn)
     def inner(self, *args, **kwargs):
+        if Dep.stack and Dep.stack[-1].readonly:
+            raise ReadonlyError()
         key = args[0]
         is_new = key not in proxy_db.attrs(self)["keydep"]
         old_value = getitem_fn(self.target, key) if not is_new else None
@@ -277,6 +281,8 @@ def delete_trap(method, obj_cls):
 
     @wraps(fn)
     def inner(self, *args, **kwargs):
+        if Dep.stack and Dep.stack[-1].readonly:
+            raise ReadonlyError()
         retval = fn(self.target, *args, **kwargs)
         proxy_db.attrs(self)["dep"].notify()
         for key in self._orphaned_keydeps():
@@ -292,6 +298,8 @@ def delete_key_trap(method, obj_cls):
 
     @wraps(fn)
     def inner(self, *args, **kwargs):
+        if Dep.stack and Dep.stack[-1].readonly:
+            raise ReadonlyError()
         retval = fn(self.target, *args, **kwargs)
         key = args[0]
         proxy_db.attrs(self)["dep"].notify()
--- a/observ/observables.py
+++ b/observ/observables.py
@@ -13,8 +13,9 @@ def observe(obj, deep=True, readonly=False):
     if not isinstance(obj, (dict, list, tuple, set)):
         return obj  # common case first
     elif isinstance(obj, dict):
+        cls = ReadonlyDictProxy if readonly else DictProxy
         if not isinstance(obj, DictProxyBase):
-            reactive = ReadonlyDictProxy(obj) if readonly else DictProxy(obj)
+            reactive = cls(obj)
         else:
             reactive = obj
         if deep:
@@ -22,8 +23,9 @@ def observe(obj, deep=True, readonly=False):
                 reactive[k] = observe(v, deep=deep, readonly=readonly)
         return reactive
     elif isinstance(obj, list):
+        cls = ReadonlyListProxy if readonly else ListProxy
         if not isinstance(obj, ListProxyBase):
-            reactive = ReadonlyListProxy(obj) if readonly else ListProxy(obj)
+            reactive = cls(obj)
         else:
             reactive = obj
         if deep:
--- a/observ/observables.py
+++ b/observ/observables.py
@@ -73,7 +73,18 @@ class ProxyDb:
         if obj_id not in self.db:
             self.register_target(proxy.target)
 
-        self.db[obj_id]["proxies"][(proxy.readonly, proxy.shallow)] = proxy
+        # Use setdefault to put the proxy in the proxies dict. If there
+        # was an existing value, it will return that instead. There shouldn't
+        # be an existing value, so we can compare the objects to see if we
+        # should raise an exception.
+        # Seems to be a tiny bit faster than checking beforehand if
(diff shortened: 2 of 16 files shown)


observ/observables.py
```diff
@@ -3,143 +3,228 @@
...
+        if self.db[obj_id]["ref"] == 0:
+            del self.db[obj_id]
+        elif self.db[obj_id]["ref"] < 0:
+            raise RuntimeError("ye olde 'this should never happen'")
+
+    def attrs(self, proxy):
+        return self.db[id(proxy.obj)]["attrs"]
+
+    def get_proxy(self, obj, readonly=False, shallow=False):
+        return next(iter(self.db[id(obj)]["proxies"][readonly][shallow]), None)
```
Korijn: Use .get

Korijn:
I would recommend adding a fixture with `autouse=True` to clear the global proxydb before every test

observ/watcher.py
```diff
@@ -14,14 +14,20 @@
...
+    that all values have been 'get'
+    """
     _traverse(obj, set())
 
 
-def _traverse(obj, seen):
+def _traverse(obj, seen: set):
     seen.add(id(obj))
-    if isinstance(obj, dict):
+    if isinstance(obj, Mapping):
```
berendkleinhaneveld, line 26: Changing `dict` to `Mapping` and `(list, tuple, set)` to `Container` was key in making the nested array reactive.

berendkleinhaneveld: So no need for a custom implementation of `KeysView` 👍

Korijn: I pushed a test to show you why you still need to trap iter methods. See `test_dict_iter`

I do think that maybe you can get away with only specializing `__iter__` since it looks like perhaps `.items()` depends on that function

berendkleinhaneveld: Good point!

Korijn: I added a couple more test cases for these since they can be a little subtle. Hope this helps.

observ/observables.py
```diff
@@ -21,57 +22,69 @@ class ProxyDb:
...
+        Creates an entry for the given target (object) and adds a strong
+        reference. The items
         """
-        obj_id = id(proxy.target)
         attrs = (
-            {"dep": Dep(), "keydep": {key: Dep() for key in dict.keys(proxy.target)}}
-            if isinstance(proxy.target, dict)
+            {
+                "dep": Dep(),
+                "keydep": {key: Dep() for key in dict.keys(target)},
```
Korijn: why not `target.keys()`?

observ/observables.py
```diff
@@ -21,57 +22,69 @@ class ProxyDb:
...
+            }
+            if isinstance(target, dict)
             else {"dep": Dep()}
```
Korijn: you could probably write this a little more succinctly without duplicating the "dep" key instantiation

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
-        if deep:
-            for k, v in reactive.items():
-                reactive[k] = observe(v)
-        return reactive
-    elif isinstance(obj, list):
-        if not isinstance(obj, ObservableList):
-            reactive = ObservableList(obj)
+            # Unpack the target from the proxy to create
+            # a new proxy further down this function
+            target = target.target
```
Korijn, line 162: can you explain this change?

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
+    # Otherwise, create a new proxy
+    proxy_type = None
+    if isinstance(target, dict):
+        proxy_type = DictProxy if not readonly else ReadonlyDictProxy
+    elif isinstance(target, list):
+        proxy_type = ListProxy if not readonly else ReadonlyListProxy
+    elif isinstance(target, set):
+        proxy_type = SetProxy if not readonly else ReadonlySetProxy
+    elif isinstance(target, tuple):
+        # FIXME: think of what needs to be done for tuple
```
Korijn: A tuple could hold dicts, lists, sets. I think that's what we need to deal with.

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
+        proxy_type = Proxy
+    return proxy_type(target, readonly=readonly, shallow=shallow)
+
+
+def reactive(target):
+    return proxy(target)
```
Korijn: please use `functools.partial` for all four of these functions, they're more efficient. As you can see in the following benchmarks, it seems to almost completely eliminate the duplicate function call overhead.

```python
import timeit

setting = """
import functools

def f(a, b, c):
    pass

def g(b, c):
    return f(4, b, c)

h = functools.partial(f, 4)
"""

print(timeit.timeit('f(4, 5, 3)', setup = setting, number=100000))
print(timeit.timeit('g(5, 3)',    setup = setting, number=100000))
print(timeit.timeit('h(5, 3)',    setup = setting, number=100000))
```

```
λ poetry run python test.py
0.007394100000000001
0.015873200000000004
0.008768300000000007
```

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
-        reactive = obj  # tuples are immutable
-        if deep:
-            reactive = tuple(observe(v) for v in reactive)
-        return reactive
-    elif isinstance(obj, set):
-        if deep:
-            return ObservableSet(observe(v) for v in obj)
+            key, value = self.iterator.__next__()
+        return (
+            proxy(key, readonly=self.readonly),
```
Korijn: when does a key need to be proxied? (can't think of any examples)

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
+        self.iterator = iterator
+        self.readonly = readonly
+
+    def __iter__(self):
+        self._iter = self.iterator.__iter__()
+        return self
+
+    def __next__(self):
+        if hasattr(self, "_iter"):
+            key, value = self._iter.__next__()  # .__next__()
```
Korijn: can you explain why this check is necessary?

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
-
-        return inner
-
-    def delete_key(fn):
-        @wraps(fn)
-        def inner(self, *args, **kwargs):
-            retval = fn(self, *args, **kwargs)
+
+            def proxied(value):
+                return proxy(value, readonly=self.readonly)
```
Korijn: use `functools.partial` !

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
-            fn = getattr(cls, name)
-            setattr(cls, name, decorate(fn))
+def write_trap(method, obj_cls):
+    fn = getattr(obj_cls, method)
 
-    return cls
+    @wraps(fn)
+    def inner(self, *args, **kwargs):
+        args = tuple(observe(a) for a in args)
+        kwargs = {k: observe(v) for k, v in kwargs.items()}
```
Korijn: `observe`? looks like this code hasn't been updated completely yet

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
+def write_key_trap(method, obj_cls):
+    fn = getattr(obj_cls, method)
+    getitem_fn = getattr(obj_cls, "__getitem__")
+
+    @wraps(fn)
+    def inner(self, *args, **kwargs):
+        key = args[0]
+        is_new = key not in proxy_db.attrs(self)["keydep"]
+        old_value = getitem_fn(self.target, key) if not is_new else None
+        args = [key] + [observe(a) for a in args[1:]]
```
Korijn: `observe`? looks like this code hasn't been updated completely yet

observ/observables.py
```diff
@@ -3,203 +3,437 @@
...
+    "KEYWRITERS": readonly_trap,
+    "DELETERS": readonly_trap,
+    "KEYDELETERS": readonly_trap,
+}
+
+
+def make_observable(proxy_cls, obj_cls, traps, trap_map):
+    for trap_type, methods in traps.items():
+        for method in methods:
+            # trap = trap_map[trap_type](method, proxy_cls, obj_cls)
```
Korijn: old code :)

observ/observables.py
```diff
@@ -290,9 +541,27 @@ class ObservableSet(set):
...
+make_observable(SetProxy, set, set_traps, trap_map)
+make_observable(ReadonlySetProxy, set, set_traps, trap_map_readonly)
+
 
-    @wraps(set.__init__)
-    def __init__(self, *args, **kwargs):
-        set.__init__(self, *args, **kwargs)
-        self.__dep__ = Dep()
+def observe(target):
+    return reactive(target=target)
```
Korijn: I would propose to drop `observe` :) users should use one of the four methods defined earlier

Korijn:
Just wondering; we should maybe add a test to see what happens if someone does this:

```python
state = reactive({"foo": 5})
state["bar"] = reactive(["something", "else"])
```

Because I believe that's legal in Vue.js, it might already just work out perfectly fine in our code... I'm not sure.

observ/observables.py
berendkleinhaneveld: Don't know anymore actually :/ I think this bit me when I tried to have a custom property on Proxy. I'll update it.

Korijn:
I guess a test for toRaw could be serializing state to json. That will fail if there are any Proxies remaining in the object being serialized.

observ/observables.py
berendkleinhaneveld: Cool, applied it!

observ/observables.py
berendkleinhaneveld: This was needed when I used custom classes for both `__iter__` and `items`. The item returned with `items` is not iterable just yet (is a `dict_items` object), so you have to call `iter()` to get a `dict_itemiterator`.
Anyways: I've replaced this whole class with a generator function instead ;)

observ/observables.py
berendkleinhaneveld: I've simply replaced this with `proxy`. Will give it some more thought tomorrow on whether I should add `readonly` and/or `shallow` arguments.

observ/observables.py
berendkleinhaneveld: Removed

observ/observables.py
```diff
@@ -2,204 +2,427 @@
...
+        iterator = fn(self.target, *args, **kwargs)
+        if self.shallow:
+            return iterator
+        if method == "items":
+
+            def proxy_dict_items(iterator, readonly=False):
+                for key, value in iterator:
+                    yield (
+                        proxy(key, readonly=readonly),
+                        proxy(value, readonly=readonly),
```
Korijn: You can pass self.readonly here from the enclosing scope to simplify :)

observ/observables.py
berendkleinhaneveld: Yes, previously there was an error in the logic because it depended on the assumption that `isinstance(our_proxy, (dict, list, tuple, set))` would result in `True`, which it doesn't. 
I've actually updated the code (together with some tests) which should make this part a bit more clear I think.

observ/observables.py
berendkleinhaneveld: Yeah, good one: dict keys need to be hashable and all the datatypes that we're supporting are not. Will update.

observ/observables.py
berendkleinhaneveld: I actually had that before and decided that I like this better :)

berendkleinhaneveld: Ok, I managed to just use a generator expression instead, so I guess it's now as compact as it needs to be :)

berendkleinhaneveld:
We (might) have a problem: because Proxy objects are hashable, we can add them to sets... When converting back to raw object, it might fail, because sets can only contain hashable types, while the underlying datatype of the proxy is *not* be hashable (dict, set and list).
For now I'll just make sure that we don't do this in a test and maybe make an example of how *not* to do it. But we probably customize the write traps for set (and maybe write key traps for dict?) to warn or fail if the user supplies a proxy object as set item or dict key.

observ/observables.py
berendkleinhaneveld: > Just wondering; we should maybe add a test to see what happens if someone does this:
> 
> ```python
> state = reactive({"foo": 5})
> state["bar"] = reactive(["something", "else"])
> ```
> 
> Because I believe that's legal in Vue.js, it might already just work out perfectly fine in our code... I'm not sure.

Added a small test and it seems to work.

Korijn:
> We (might) have a problem: because Proxy objects are hashable, we can add them to sets... When converting back to raw object, it might fail, because sets can only contain hashable types, while the underlying datatype of the proxy is _not_ be hashable (dict, set and list). For now I'll just make sure that we don't do this in a test and maybe make an example of how _not_ to do it. But we probably customize the write traps for set (and maybe write key traps for dict?) to warn or fail if the user supplies a proxy object as set item or dict key.

Maybe this is stupid but I think they shouldn't be hashable. Right? They don't define `__hash__` so they aren't hashable, I think. :/

If not, maybe we can figure out how to _make_ them unhashable. It doesn't make sense for them to be hashable.

berendkleinhaneveld:
> > We (might) have a problem: because Proxy objects are hashable, we can add them to sets... When converting back to raw object, it might fail, because sets can only contain hashable types, while the underlying datatype of the proxy is _not_ be hashable (dict, set and list). For now I'll just make sure that we don't do this in a test and maybe make an example of how _not_ to do it. But we probably customize the write traps for set (and maybe write key traps for dict?) to warn or fail if the user supplies a proxy object as set item or dict key.
> 
> Maybe this is stupid but I think they shouldn't be hashable. Right? They don't define `__hash__` so they aren't hashable, I think. :/
> 
> If not, maybe we can figure out how to _make_ them unhashable. It doesn't make sense for them to be hashable.

Haha, well, I tried and succeeded already in making them non-hashable (by setting the `__hash__` class method to `None`). But the problem I got was: how do we manage Proxies in the `proxy_db`? Because there, we use a `WeakSet` for holding the proxies. But if they are not hashable, then we would need some other way of keeping a weak reference. 

Now that I'm writing this: maybe we could use a `weakref.WeakValueDictionary` instead to keep weak references to Proxies around. 🤔 

Edit: sweet, I actually got it working! Coming up in the next commit!

observ/observables.py
```diff
@@ -239,6 +239,8 @@ def write_trap(method, obj_cls):
 
     @wraps(fn)
     def inner(self, *args, **kwargs):
+        if Dep.stack and Dep.stack[-1].readonly:
```
berendkleinhaneveld: I'm not too sure about this, but it seems to do the trick for figuring out whether a computed (readonly) expression is now being run. What do you think?
The goal is to prevent side-effects (and infinite recursive loops) from happening for computed expressions.

Korijn: Is ` and Dep.stack[-1].readonly` even necessary? :O

I mean; would this not be generally desirable, even outside computed? To prevent users from changing state while a watcher expression is being evaluated seems pretty valuable; it's only allowed in callbacks. So the readonly kwarg on watcher is kind of redundant then. I guess the problem is sync callbacks?

Maybe this should be kept for a separate PR... 

I do think it should have a clearly distinguishable exception type, if possible

berendkleinhaneveld: Good question: do we want to allow changes to state for watched expressions? If not, then we can simplify this :)

berendkleinhaneveld (commented):
Good point about a separate exception.
Also, the commit message could ve better: 'don't allow modification of state in watched expressions'

observ/observables.py
```diff
@@ -2,204 +2,443 @@
...
+        if obj_id not in self.db:
+            # When there are failing tests, it might happen that proxies
+            # are garbage collected at a point where the proxy_db is already
+            # cleared. That's why we need this check here.
+            # See fixture [clear_proxy_db](/tests/conftest.py:clear_proxy_db)
+            # for more info.
+            return
+        self.db[obj_id]["refs"] -= 1
+
+        if self.db[obj_id]["refs"] <= 0:
```
Korijn: If WeakValueDictionary supports `len()` you can use that instead of the reference counter!

observ/observables.py
```diff
@@ -2,204 +2,443 @@
...
+    def reference(self, proxy):
+        """
+        Adds a reference to the collection for the wrapped object's id
+        """
+        obj_id = id(proxy.target)
+
+        if obj_id not in self.db:
+            self.register_target(proxy.target)
+
+        self.db[obj_id]["proxies"][(proxy.readonly, proxy.shallow)] = proxy
```
Korijn: Should this raise an error if the key is already populated?

observ/observables.py
```diff
@@ -2,204 +2,443 @@
...
+    "KEYREADERS": read_key_trap,
+    "ITERATORS": iterate_trap,
+    "WRITERS": readonly_trap,
+    "KEYWRITERS": readonly_trap,
+    "DELETERS": readonly_trap,
+    "KEYDELETERS": readonly_trap,
+}
+
+
+def make_reactive(proxy_cls, obj_cls, traps, trap_map):
```
Korijn: I was thinking of renaming this to `bind_traps` to clearly distinguish it from `reactive`, what do you think?

observ/api.py
```diff
@@ -4,12 +4,12 @@
...
 
 from .dep import Dep
-from .observables import observe
+from .observables import reactive, to_raw
 from .scheduler import scheduler
 from .watcher import Watcher
 
 
-__all__ = ("observe", "computed", "watch", "scheduler")
+__all__ = ("reactive", "computed", "watch", "scheduler", "to_raw")
```
Korijn: I think this should also include shallow, readonly and readonly_shallow

observ/observables.py
berendkleinhaneveld: Yes, I also had problems with the name and I like the suggestion!

observ/observables.py
berendkleinhaneveld: Yes! I had that as a fleeting thought but forgot after successful implementation :')

observ/api.py
berendkleinhaneveld: Good one. I've actually put them now in api.py next to computed / watch. If you're not ok with that, let me know, otherwise resolve this comment

observ/observables.py
berendkleinhaneveld: Well, that's going to be hard, because the proxy will only be dereferenced _after_ this method. So within this method that is not going to work. Although we might check to see if the given proxy is the last referenced proxy 🤔

Korijn: Good point!

berendkleinhaneveld: Nice, that actually works!

observ/observables.py
berendkleinhaneveld: Ok, it's good as it is!

PR #36, opened 2021-10-25
Fix update and __ior__ methods of dict

berendkleinhaneveld:
Apparently `update` and `__ior__` were broken, because no dependencies for updated keys were notified.

I added a check in the write_trap to see if it is a dict, and then figure out which keys are affected. Might instead just create a different `write_trap` function just for dict.

### DIFF
--- a/observ/observables.py
+++ b/observ/observables.py
@@ -253,9 +253,8 @@ def write_trap(method, obj_cls):
             raise StateModifiedError()
         args = tuple(proxy(a) for a in args)
         kwargs = {k: proxy(v) for k, v in kwargs.items()}
+        retval = fn(self.target, *args, **kwargs)
 
-        old_keys = set()
-        updated_keys = set()
         if obj_cls == dict:
             """
             From the docs about `update` (__ior__ just accepts other dicts):
@@ -264,7 +263,6 @@ def write_trap(method, obj_cls):
             If keyword arguments are specified, the dictionary is then updated
             with those key/value pairs: d.update(red=1, blue=2).
             """
-            old_keys = set(self.target.keys())
             if args:
                 if hasattr(args[0], "keys"):
                     # First argument is (wrapped) dict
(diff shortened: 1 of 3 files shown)


observ/observables.py
```diff
@@ -253,7 +253,37 @@ def inner(self, *args, **kwargs):
             raise StateModifiedError()
         args = tuple(proxy(a) for a in args)
         kwargs = {k: proxy(v) for k, v in kwargs.items()}
+
+        old_keys = set()
+        updated_keys = set()
```
Korijn: You can put these in `else` to shave off two redundant set allocations

observ/observables.py
```diff
@@ -253,7 +253,37 @@ def inner(self, *args, **kwargs):
...
+                # Key/value pairs as keyword arguments
+                updated_keys = set(kwargs.keys())
         retval = fn(self.target, *args, **kwargs)
+        new_keys = updated_keys - old_keys
+        updated_keys &= old_keys
+
+        for key in new_keys:
+            proxy_db.attrs(self)["keydep"][key] = Dep()
+        for key in updated_keys:
+            proxy_db.attrs(self)["keydep"][key].notify()
```
Korijn: Shouldn't all of this go into the if statement?

berendkleinhaneveld: No, not all of it. There is a split, because the set with old keys needs to be created before the function `fn` is called and the keydeps need to be called after the function is called.

berendkleinhaneveld: Or, I have to check in the keydeps if there is an existing dep.

Korijn: Ah I see, makes sense

berendkleinhaneveld: I've now reworked it anyways. Haven't taken the time to check performance difference with my first attempt, but I guess we'll dive into that another time (when we have an actual benchmark setup).

observ/observables.py
```diff
@@ -254,6 +254,33 @@ def inner(self, *args, **kwargs):
...
+                    # First argument is (wrapped) dict
+                    updated_keys = set(args[0].keys())
+                else:
+                    # First argument is iterable of key/value pairs
+                    updated_keys = set(key for key, _ in args[0])
+            else:
+                # Key/value pairs as keyword arguments
+                updated_keys = set(kwargs.keys())
+
+            keydeps = proxy_db.attrs(self)["keydep"]
```
Korijn: you can get a reference to `proxy_db.attrs(self)` before the if-statement and then reuse the reference on line 285 to save a double lookup

PR #39, opened 2021-10-27
Don't notify when there is no change

berendkleinhaneveld:


observ/observables.py
```diff
@@ -253,36 +253,32 @@ def inner(self, *args, **kwargs):
...
-
+            change_detected = False
             keydeps = attrs["keydep"]
-            for key in updated_keys:
-                if key in keydeps:
-                    keydeps[key].notify()
-                else:
-                    keydeps[key] = Dep()
+            for key, val in self.target.items():
+                if ids.get(id(key)) != id(val):
```
Korijn: you can use the `is` operator to test if two variables point to the same object in memory, that might simplify things

berendkleinhaneveld: I think that won't work, because the original object might have been swapped and because we're dealing with references here, we will be comparing the same (updated) object, which will always result in a True comparison. That's why I build a new list/dict with the ids, so I know for sure they won't change.

Korijn: Well, I mean, you can create a shallow copy which is presumably more efficient and simpler to do.

Instead of:

```
    if obj_cls == dict:
        # Create dict of the ids of all key and values in the dict
        ids = {id(key): id(val) for key, val in self.target.items()}
    else:  # list and set
        # Create list of the ids of all items in the list
        ids = list(id(item) for item in self.target)
```

You can do:

```python
old = self.target.copy()  # shallow copy, no need to distinguish types
```

Then you can write:

```python
for key, val in self.target.items():
    if old.get(key) is not val:
```

And:

```python
    if old != self.target:
        attrs["dep"].notify()
```

Which seems a little less finnicky to me. But I might be glossing over more details.

berendkleinhaneveld: Ah, right: shallow copy! That should work fine indeed!

observ/observables.py
```diff
@@ -253,36 +253,25 @@ def inner(self, *args, **kwargs):
...
-                else:
-                    keydeps[key] = Dep()
+            for key, val in self.target.items():
+                if old.get(key) is not val:
+                    if key in keydeps:
+                        keydeps[key].notify()
+                    else:
+                        keydeps[key] = Dep()
+                    change_detected = True
+                if change_detected:
```
Korijn: shouldn't this be outside the for loop?

berendkleinhaneveld: 🤦‍♂️ yep, good catch!

Korijn:
LGTM!

PR #41, opened 2021-11-01
Store with undo and redo

berendkleinhaneveld:
This adds a `Store` class which facilitates keeping track of state and providing undo/redo support.

Client applications can subclass the `Store` and define mutations and computed props on this class. The current state of the `Store` is accessible through a reactive (readonly) `state` property on `Store`. Other useful methods are `undo`, `redo`, `can_undo` and `can_redo` (those last two are actually reactive).


observ/store.py
```diff
@@ -0,0 +1,135 @@
...
+
+registry = {}
+
+
+def mutation(fn: T) -> T:
+    registry[fn] = "mutation"
+
+    @wraps(fn)
+    def inner(self, *args, **kwargs):
+        return fn(self, *args, **kwargs)
```
Korijn: If you're not altering the behavior in any way, you can skip all this and just `return fn` as is!

observ/store.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def computed(fn: T) -> T:
+    parameters = signature(fn).parameters
+    if len(parameters) == 1 and "self" in parameters:
+        registry[fn] = "computed"
+
+        @wraps(fn)
+        def inner(self):
+            return fn(self)
```
Korijn: I think you can just `return fn` here as well

observ/store.py
```diff
@@ -0,0 +1,135 @@
...
+        @wraps(fn)
+        def getter():
+            if watcher.dirty:
+                watcher.evaluate()
+            if Dep.stack:
+                watcher.depend()
+            return watcher.value
+
+        getter.__watcher__ = watcher
+        return getter
```
Korijn: This is confusing, why not just `return computed(fn)`?

observ/store.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, state):
+        self._present = reactive(state)
+        # Because we can't work with diff patches (yet) we'll have to copy
+        # over the whole state. When triggering undo/redo, we'll therefore
+        # trigger all the watchers / computed expressions...
+        self._past = shallow_reactive([copy.deepcopy(to_raw(state))])
+        self._future = shallow_reactive([])
+        self.state = readonly(state)
+
+        computed_props = registered_computed_props.setdefault(id(self), {})
```
Korijn: This can be a local dict on self right? No need for a global store or anything

observ/store.py
```diff
@@ -0,0 +1,135 @@
...
+        for method_name in dir(self):
+            method = getattr(self, method_name)
+            if not hasattr(method, "__wrapped__"):
```
Korijn: Why not check if the method is in `registry`? Seems more appropriate!

observ/store.py
berendkleinhaneveld: I should have reviewed the code a bit better before pushing I guess :)

observ/store.py
berendkleinhaneveld: That's way more simple!

observ/store.py
berendkleinhaneveld: Because `computed` *is* now this function. A bit confusing, yes. I might split it up again, but then I'll probably have to rename one of them.

Korijn: That.. is too complicated :) let's think of something to make it nicer

observ/store.py
berendkleinhaneveld: Just looked at the code and remembered why I did this. We override `__getattribute__` and if you get a local prop on `self`, you will end up in an endless loop. So that's why I stored it outside the object, in order to circumvent that. I can see if there is another way of dealing with the endless loop.

berendkleinhaneveld: Ok, I can just check for the name of the local dict.

Korijn: You can grab it off `super()` I think, or otherwise use `super().__getattribute__("name of your dict")` to avoid hitting your implementation of `__getattribute__`

observ/store.py
berendkleinhaneveld: Ok, I came up with the following: if we don't expose the `store` through the api, then there doesn't need to be a name clash. But in order to prevent circular imports I've moved the original definition of `computed` to [/observ/watcher.py]()

Because the store is a pretty specific piece of code, I think it would be okay to leave it out of the default api. If someone wants to implement a subclass of the `Store`, then the following imports would suffice:
```python
from observ.store import computed, mutation, Store
```

observ/api.py
```diff
@@ -21,21 +24,6 @@
...
-            watcher.evaluate()
-        if Dep.stack:
-            watcher.depend()
-        return watcher.value
-
-    getter.__watcher__ = watcher
-    return getter
-
-
 def watch(fn, callback, sync=False, deep=False, immediate=False):
```
berendkleinhaneveld: Maybe for good measure I could move this to [observ/watcher.py]() as well

observ/watcher.py
```diff
@@ -176,3 +177,18 @@ def depend(self) -> None:
     @property
     def fn_fqn(self) -> str:
         return f"{self.fn.__module__}.{self.fn.__qualname__}"
+
+
+def computed(fn):
```
berendkleinhaneveld: Add type annotation

observ/store.py
```diff
@@ -0,0 +1,108 @@
...
+                )
+
+    def __getattribute__(self, name):
+        if name != "_computed_props":
+            fn = self._computed_props.get(name)
+            if fn:
+                return fn()
+        return super().__getattribute__(name)
+
+    def commit(self, fn, *args, **kwargs):
```
berendkleinhaneveld: Add type annotation

observ/store.py
```diff
@@ -0,0 +1,108 @@
...
+    registry[fn] = "computed"
+    return fn
+
+
+class Store:
+    """
+    Store that tracks mutations to state in order to enable undo/redo functionality
+    """
+
+    def __init__(self, state):
```
berendkleinhaneveld: Add type annotation / doc

PR #42, opened 2021-11-02
Refactor the store to make mutations nicer

berendkleinhaneveld:
Instead of:
```python
class MyStore(Store):
    @mutation
    def add(self, state, amount):
        state["counter"] += amount
```
we can do:
```python
class MyStore(Store):
    @mutation
    def add(self, amount):
        self.state["counter"] += amount
```
So mutations look and read more like proper functions.

### DIFF
--- a/examples/store_with_undo_redo.py
+++ b/examples/store_with_undo_redo.py
@@ -8,20 +8,20 @@ from observ.store import computed, mutation, Store
 
 class CounterStore(Store):
     @mutation
-    def bump_count(self, state):
+    def bump_count(self):
         """
         Bump counter by one.
 
-        Note: don't pass in the state argument: that is taken
-        care of by the store. By decorating it with `mutation`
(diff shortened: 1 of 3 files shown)


examples/store_with_undo_redo.py
```diff
@@ -8,20 +8,20 @@
...
         """
         Bump counter by one.
 
-        Note: don't pass in the state argument: that is taken
-        care of by the store. By decorating it with `mutation`
-        the store wraps this method and passes in a writable
-        version of the state.
+        Note: normally self.state is a readonly proxy on the present
+        state, but because this method is decorated with `mutation`
+        `self.state` is replaced with the mutatable `self._present`
```
Korijn: I'm picturing a mutant table in my head now. It's `mutable` :)

Korijn:
You can also remove the comments like `# Note that no state argument is provided here!` from the example now :)

PR #44, opened 2021-11-02
Add deep argument to computed decorators

berendkleinhaneveld:
Now it's possible to create deep computed expressions to enable deep watching on data structures.


Korijn:
LGTM!!

PR #47, opened 2021-11-02
Fix real deep watchers

berendkleinhaneveld:
I had a problem with some changes deep within some structures not coming through.
Apparently, a `DictProxy` is *not* an instance of `Mapping`... And the same goes for the other proxies. I thought I had tried and tested this before (in a REPL I believe), but I must have done something weird.

Anyways, I've updated `traverse` with checking for the specific types that we support and now it all seems to work as intended!


Korijn:
LGTM

PR #50, opened 2021-11-08
Implement patch-based undo redo in the store

berendkleinhaneveld:


observ/store.py
```diff
@@ -20,8 +21,10 @@ def inner(self, *args, **kwargs):
         readonly_state = self.state
         self.state = self._present
         try:
+            current = to_raw(self.state)
```
Korijn, line 24: I think this still requires a deepcopy (but only here)

berendkleinhaneveld: I was doubting it, because actually the to_raw function already recursively calls copy on all enclosed data objects. So I guess that works kind of like deep copy already. 
Otherwise: I would love to see if we can come up with a test that verifies that a deepcopy is needed. Gotta give it some more thought.

berendkleinhaneveld: to_raw does not make copies of sets! So I could either add that call there or create a deepcopy. I think creating a shallow copy of set would be better

Korijn: Hope you don't mind me butting in here :) I made the change you proposed and optimized to_raw a little on the way.

I still think we should add that test you mentioned.

berendkleinhaneveld: I just added the test! I think we're good to go now.

PR #52, opened 2021-11-09
Make computed deep by default

### LINKED ISSUE #49: Computed deep=True default
I've been thinking and it just makes more sense to have deep=True as a default on computed

berendkleinhaneveld:
Fixes #49.

In watcher.py I had to adjust the try/finally statement because when there is an exception, `value` will not be defined, so another exception will be thrown. And in the tests, we trigger such an exception (StateModifiedError).

### DIFF
--- a/observ/store.py
+++ b/observ/store.py
@@ -32,7 +32,7 @@ def mutation(fn: T) -> T:
     return inner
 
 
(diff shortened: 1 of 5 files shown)


Korijn:
LGTM!

PR #56, opened 2021-11-12
Make sure proxy.target does not contain proxies

berendkleinhaneveld:
This should make the `to_raw` function much simpler.

The `to_raw` function does not do a deep check, so if values are assigned that have 'deep' proxied objects, then that's going to be a problem. So we might need a deep `to_raw` conversion function anyways. We could pass it as an argument for the `to_raw` function.


berendkleinhaneveld:
Ok, something weird happened here. :/

PR #57, opened 2021-11-12
Remove unnecessary calls to proxy from write traps

berendkleinhaneveld:
This PR now simply removes the proxy calls from the write traps, to save a bit of overhead.


Korijn:
The changes look good and appear to match what is done in Vue 3. I still wonder what would happen if a user inserts a Proxy deep inside nested stuctures. I'll try to add a test to see what happens.

berendkleinhaneveld:
Yeah, exactly: that will most likely break it. Looking forward to the test 👍

Korijn:
You've got this change right, so feel free to merge. 🚀

observ/store.py
```diff
@@ -21,7 +22,7 @@ def inner(self, *args, **kwargs):
         readonly_state = self.state
         self.state = self._present
         try:
-            current = to_raw(self.state)
+            current = deepcopy(to_raw(self.state))
             fn(self, *args, **kwargs)
             ops, reverse_ops = patchdiff.diff(current, self.state)
```
Korijn: you might want to check/test how this behaves in the problem case described in the xfailing test. Might be necessary to add a recursive to_raw (the old behaviour) as well...

I should say; Vue also has this "problem"(?), and doesn't seem to handle it any better than we do. So... yeah. You are probably best off just avoiding it (instead of using recursive to_raw)

berendkleinhaneveld: yeah, I'm leaning towards removing the to_raw from the write traps and just restore the to_raw function. 
Because within our traps it doesn't really matter apparently, and the to_raw function will then ensure a non-proxied object if needed (such as in the store). And then I'll remove the deepcopy again from the store.

Korijn:
We'll find out soon enough how well this behaves I guess! :)

PR #62, opened 2021-11-26
Benchmark

### LINKED ISSUE #38: Benchmarks
Would be nice to have a benchmark setup so that we can make informed decisions on performance.

We could leverage pytest-benchmark together with [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark).
And naturally we would need to have some proper functions to run as benchmark.

Korijn:
This is just one way to do benchmarking. Supposedly this makes the benchmark consistent across machines and time. See https://pythonspeed.com/articles/consistent-benchmarking-in-ci/

The downside is that this doesn't provide any sort of "benchmark suite" mechanism like pytest-benchmark would.

Also I am a little unsure of how to track performance changes over time, ideally we can graph them as PRs are merged, somehow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,29 @@ jobs:
       - name: Test
         run: poetry run pytest --cov=observ --cov-report=term-missing
 
+  bench:
+    name: Benchmark ${{ matrix.name }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install poetry
+        run: pip install "poetry>=1.1.8,<1.2"
+      - name: Install dependencies
+        run: |
+          poetry install
+      - name: Install valgrind
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -qq valgrind
+      - name: Benchmark
+        run: |
+          export PYTHONHASHSEED=1234
+          poetry run python bench/cachegrind.py python bench/suite.py
+
   build:
     name: Build and test wheel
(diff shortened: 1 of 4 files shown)


berendkleinhaneveld:
Cool article, nice idea about tracking CPU instruction count!

I don't fully grasp what you mean by 'benchmark suite': the ability to track differences through changes and compare them? Or integration with an existing test suite? 

We could track performance metrics by storing them as build artifacts (90 days retention). But the retention limit is not that great... We could also create an 'empty' branch, and commit results of every build (keyed on commit id) to a (json) file. A bit hacky/wacky, but that would give us proper retention ;)

Or a tiny bit less hacky: we could use GitHub CLI within GitHub Actions to create/edit a gist where we can append benchmark results for each build. That way the benchmark data don't become part of the repository.

Korijn:
> I don't fully grasp what you mean by 'benchmark suite': the ability to track differences through changes and compare them? Or integration with an existing test suite?

Mostly to provide a framework; ideally we can define various benchmark routines (e.g. we might want to separately benchmark read and write performance, or even separately per datastructure type, etc), and have a cli command that runs and analyzes the routines, and provides utilities to report and track the output data.

With the approach here we would have to set all that up ourselves. Pytest-benchmarks solves a big chunk of that work.

Korijn:
I've also been thinking that it's perhaps most interesting to design benchmarks that compare mutations on raw datastructures and proxied datastructures. Then we can use the overhead ratio as a metric, and see how that evolves over time.

Korijn:
Pytest benchmark just looks a lot better. I'm still figuring out how to setup the comparison properly (between raw and proxied)

Korijn:
This seems to do exactly what I want: https://github.com/psf/pyperf

PR #65, opened 2022-02-21
Tryout to generate patchdiff ops on-the-fly

berendkleinhaneveld:
I've taken a very naive and explorative approach to generating patchdiff ops. There are tons of problems with this, but I guess that that is exactly what I wanted to figure out: what would we want from the generated ops and how are we going to achieve that?

I started out with adding another argument `ops` to the callback function, so that you can receive the actual patchdiff ops next to the `new` and `old` value. Already the first problem with this is: within the watcher expression, there could be any type of expression, while we can only easily generate `ops` for the changes in the underlying reactive data structures. So the `ops` returned here might not map directly to the `old` and `new` values...

Then there is another problem: how do you keep track of the 'paths' for the `ops`?

### DIFF
--- a/observ/dep.py
+++ b/observ/dep.py
@@ -22,6 +22,29 @@ class Dep:
         if self.stack:
             self.stack[-1].add_dep(self)
 
-    def notify(self) -> None:
+    def notify(self, ops=None) -> None:
         for sub in sorted(self._subs, key=lambda s: s.id):
-            sub.update()
+            sub.update(ops)
+
+
+class Path:
+    stack = []
+
+    @classmethod
+    def put(cls, obj, component):
+        # Maybe try to start building the path from the
+        # object that is returned in the lambda? Is that
+        # even possible?
+        if (id(obj), component) in cls.stack:
+            # Should maybe clear from that index?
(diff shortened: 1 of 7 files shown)


Korijn:
I would propose solving this with an independent set of proxies, instead of mixing the implementations. The purpose is subtly different and requires a different tracking approach as you found out.

Take some inspiration from [Immer.js](https://immerjs.github.io/immer/), provides a single API `produce`, which takes a state object and a mutator function. The object is shallowly converted to a proxy, and on every "get", the resulting object is lazily converted to a proxy as well, and a global stack is updated to track the current path. On every "set" you emit a patch object.

Immer.js does some more things, as it also returns the mutated object, but the original object is kept intact, without making a deepcopy ("structural sharing"). It probably means you need to shallowly clone collections (and all their parents) right before they are mutated via a "set", and track which parts have been replaced already so you don't redundantly do that over and over again.

Korijn:
Thinking about it some more, you have to let the mutator function alter the structure (you can't get away with only emitting patch objects). So you have to do the full blown implementation as described above in one go, I think.

PR #71, opened 2022-04-19
Fix bug with many watchers

### LINKED ISSUE #70: Add test with multiple watchers in combination with the Qt scheduler

berendkleinhaneveld:
Closes #70.

There is a bug in the traverse code that is triggered when using many watchers (on the same data?). It seems that the traverse code skips objects that have been 'seen' before, but apparently this does not seem to work how we thought it would work...

Instead, when we simply traverse everything, then all the watchers are properly queued and triggered.

### DIFF
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -71,7 +71,7 @@ class Scheduler:
             self.has.discard(watcher.id)
             watcher.run()
 
-            if self.detect_cycles:  # and watcher.id in self.has:
+            if self.detect_cycles:
                 self.circular[watcher.id] += 1
                 if self.circular[watcher.id] > 100:
                     raise RecursionError(
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -76,14 +76,16 @@ def traverse(obj, seen=None):
     # traversal) is an option but way more expensive than just using a list
     if seen is None:
         seen = []
-    seen.append(obj)
+    # since proxies may be garbage collected during traversal, it's not safe to use
+    # them for tracking, so we try to unwrap proxies where possible
+    obj_unwrapped = getattr(obj, "target", obj)
+    seen.append(obj_unwrapped)
     for v in val_iter:
         if (
             isinstance(
                 v, (dict, DictProxyBase, list, ListProxyBase, set, SetProxyBase, tuple)
             )
-            and v not in seen
-        ):
+        ) and getattr(v, "target", v) not in seen:
             traverse(v)
 
 
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -78,8 +78,7 @@ def traverse(obj, seen=None):
         seen = []
     # since proxies may be garbage collected during traversal, it's not safe to use
     # them for tracking, so we try to unwrap proxies where possible
-    obj_unwrapped = getattr(obj, "target", obj)
-    seen.append(obj_unwrapped)
+    seen.append(getattr(obj, "target", obj))
     for v in val_iter:
         if (
             isinstance(
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -72,7 +72,9 @@ def _traverse(obj, seen: set):
     else:
         return
     for v in val_iter:
-        if isinstance(v, Container) and id(v) not in seen:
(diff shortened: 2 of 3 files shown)


Korijn:
Well, first of all, this opens up the possibility of an infinite recursion cycle, no?

```py
a = {}
b = {}
a["b"] = b
b["a"] = a
```

And, other than that, I don't understand why this fixes the problem. It's not like the different objects contained in the state can have the same id, right?

berendkleinhaneveld:
Let me start by saying that I am probably as confused as you are. I also know that infinite recursion would be a problem, but I guess we could try and solve that in some other way (like how we count cycles in `scheduler.flush`)?

I am just not sure that we understand `id()` well enough in this context. It might also have to do something with the recursive nature of the `watcher.get` function.

Btw: if you want to investigate yourself, you can change the number from 100 to 3 in the test (and revert the changes to `traverse`, you should be able to reliably trigger the problem. The weird thing is that the number of watchers that do get called is not stable: it fluctuates between 0 and 2 (when you change the amount of watchers to 3). This is what confuses me the most :/

Korijn:
I'll give it a try right now.

Korijn:
> Let me start by saying that I am probably as confused as you are. I also know that infinite recursion would be a problem, but I guess we could try and solve that in some other way (like how we count cycles in `scheduler.flush`)?

The scheduler.flush method is feasible because it can index watchers by id. I thought about just tracking the recursion depth, but that wouldn't provide any advantage over the recursion error Python would already trigger on its own.

That's when I realized, we're not doing this just to detect cycles, we're doing this to actually _support_ traversal (and thereby watching) of datastructures with cycles.

> I am just not sure that we understand `id()` well enough in this context. It might also have to do something with the recursive nature of the `watcher.get` function.

Well, there is a gotcha with `id()`, from the docs: "Two objects with non-overlapping lifetimes may have the same `id()` value." So we _must_ be falling into that trap here. From that I am going to guess that our proxies are being garbage collected and recreated. This makes sense to me and explains your observations!

---

Anyway: I think I have fixed the problem by reintroducing `seen`, but with a list.

I'm wondering if it would be safer to unwrap proxies during traversal (grabbing `proxy.target`).

Korijn:
> Let me start by saying that I am probably as confused as you are. I also know that infinite recursion would be a problem, but I guess we could try and solve that in some other way (like how we count cycles in `scheduler.flush`)?

The scheduler.flush method is feasible because it can index watchers by id. I thought about just tracking the recursion depth, but that wouldn't provide any advantage over the recursion error Python would already trigger on its own.

That's when I realized, we're not doing this just to detect cycles, we're doing this to actually _support_ traversal (and thereby watching) of datastructures with cycles.

> I am just not sure that we understand `id()` well enough in this context. It might also have to do something with the recursive nature of the `watcher.get` function.

Well, there is a gotcha with `id()`, from the docs: "Two objects with non-overlapping lifetimes may have the same `id()` value." So we _must_ be falling into that trap here. From that I am going to guess that our proxies are being garbage collected and recreated. This makes sense to me and explains your observations!

---

Anyway: I think I have fixed the problem by reintroducing `seen`, but with a list.

Korijn:
> I'm wondering if it would be safer to unwrap proxies during traversal (grabbing `proxy.target`).

I'm pretty sure it is, by the logic above. Added that, tests still seem to pass!

Korijn:
I was fooling myself, forgot to pass seen down the recursion. Still works though, thankfully :) Also tried to test with much larger (10k) and much smaller (3) amounts of watchers.

Korijn:
Ah, I just realized, unwrapping is not needed. By tracking proxies in the seen list, they cannot be garbage collected during traversal! You can go ahead and merge this now :)

berendkleinhaneveld:
> Well, there is a gotcha with `id()`, from the docs: "Two objects with non-overlapping lifetimes may have the same `id()` value." So we _must_ be falling into that trap here. From that I am going to guess that our proxies are being garbage collected and recreated. This makes sense to me and explains your observations!

Aha! That's good to know!

Good idea to track the objects themselves instead of their ids.

PR #77, opened 2022-05-31
Fix for getting a non-existing key from a dict

berendkleinhaneveld:
- [x] Add failing test for getting a non-existing key from dict
- [x] Implement fix for the test

### DIFF
--- a/tests/test_usage.py
+++ b/tests/test_usage.py
@@ -625,3 +625,24 @@ def test_usage_class_instances():
     # write to a class attribute
     a[2].foo = 10
     assert called == 2  # class instances are NOT reactive
+
+
+@pytest.mark.xfail
+def test_watch_get_non_existing():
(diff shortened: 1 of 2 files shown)


tests/test_usage.py
```diff
@@ -625,3 +625,45 @@ def _callback():
...
+def test_watch_get_non_existing():
+    a = reactive({})
+
+    def result():
+        return a.get("foo", None)
+
+    watcher = watch(
+        result,
+        sync=True,
+        deep=True,
```
Korijn: I think this should not need `deep`?

berendkleinhaneveld:
Note that `"{key}" in ...` for lists and sets worked fine already: just added some tests to be sure. At first I thought they were broken as well, but that was just an error in the test 😅

Korijn (approved):

PR #86, opened 2022-09-06
Add strict argument to store for ignoring empty mutations on the store

berendkleinhaneveld:
Currently, when a mutation on the store is called that doesn't actually change anything, the undo stack is still updated.
This PR adds a `strict` argument to the Store constructor to control its behaviour:

* `strict = True` (default) will raise a `RuntimeError` when no change is made within a mutation
* `strict = False` will ignore empty mutations

### DIFF
--- a/observ/store.py
+++ b/observ/store.py
@@ -24,8 +24,11 @@ def mutation(fn: T) -> T:
             current = to_raw(self.state)
             fn(self, *args, **kwargs)
             ops, reverse_ops = patchdiff.diff(current, self.state)
-            self._past.append((ops, reverse_ops))
-            self._future.clear()
+            # If ops and reverse_ops are empty, that means
+            # that there are no actual changes to record
+            if ops or reverse_ops:
+                self._past.append((ops, reverse_ops))
+                self._future.clear()
         finally:
             self.state = readonly_state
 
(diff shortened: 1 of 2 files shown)


observ/store.py
```diff
@@ -24,8 +24,11 @@ def inner(self, *args, **kwargs):
             current = to_raw(self.state)
             fn(self, *args, **kwargs)
             ops, reverse_ops = patchdiff.diff(current, self.state)
-            self._past.append((ops, reverse_ops))
-            self._future.clear()
+            # If ops and reverse_ops are empty, that means
+            # that there are no actual changes to record
+            if ops or reverse_ops:
+                self._past.append((ops, reverse_ops))
+                self._future.clear()
```
Korijn, line 36: Personally I would consider it a bug and expect an exception if I call a mutation that doesn't change anything. But I can see a use case for preferring it to silently do nothing.

I propose adding a `strict=True` parameter to the `Store` constructor to toggle this behaviour. I think users should consciously opt into silent ignores of an edge case situation.

Korijn: No exception?

berendkleinhaneveld: Ah, right: forgot!

berendkleinhaneveld: Fixed!

Korijn (approved):

PR #90, opened 2023-10-02
Support watching of (collections of) reactive objects

### LINKED ISSUE #81: Generate watch expressions from (arrays of) watchable objects
See this docs page: https://vuejs.org/api/reactivity-core.html#watch

Vue's watch accepts four types for the first argument:

> * A getter function that returns a value
> * A ref
> * A reactive object
> * ...or an array of the above.

Right now in observ, we only support the first case.

This issue represents the need to implement support for the other three cases. Of course, case number two does not apply to observ since we don't have refs, only reactives.

In other words, allow the following type of constructions:

```python
state = reactive({"text": "Hello, world!})

def on_change(state):
    label.setText(state["text"])

watch(state, on_change)
# instead of watch(lambda: state, on_change, deep=True)
```

berendkleinhaneveld:
This allows for:

```python
a = reactive({"foo": "foo"})

watcher = watch(a, cb, ...)
```
and
```python
a = reactive({"foo": "foo"})
b = reactive(["bar"])

watcher = watch([a, b], cb, ...)
```

instead of:
```python
watcher = watch(lambda: a, cb, ...)
```


observ/watcher.py
```diff
@@ -112,7 +123,10 @@ def __init__(
         callback: Method to call when value has changed
         """
         self.id = next(_ids)
-        self.fn = fn
+        if isinstance(fn, Proxy):
+            self.fn = lambda: fn
```
Korijn: Additionally, deep should be True by default for a reactive object

observ/watcher.py
```diff
@@ -123,10 +123,13 @@ def __init__(
...
-        if isinstance(fn, Proxy):
-            self.fn = lambda: fn
-        else:
+        if callable(fn):
             self.fn = fn
+        else:
+            self.fn = lambda: fn
+            # When watching proxies or a list of proxies
+            # make sure that deep watching is enabled
+            deep = True
```
Korijn: The default should be True (in this case), but now it's impossible to set it to False for the user.

So maybe set the actual default in the code to None, so you can detect if the user is setting it or not?

tests/test_usage.py
```diff
@@ -750,11 +750,33 @@ def test_watch_reactive_object():
...
     assert cb.call_count == 1
+
+
+def test_watch_list_of_reactive_objects():
+    a = reactive({"foo": "foo"})
+    b = reactive(["bar"])
+    cb = Mock()
+
+    watcher = watch(  # noqa: F841
+        [a, b],
```
Korijn, line 768: I can't remember if the list is converted to a proxy as well. Is it? Does it matter?

Edit: I guess not, the list is not going to change.

observ/watcher.py
berendkleinhaneveld: Ah, indeed. Good point about setting the default value to None

Korijn (approved):

PR #92, opened 2023-10-23
Drop python 3.7 and add CI for 3.11 and 3.12

berendkleinhaneveld:
Bonus: update poetry to 1.4.2


pyproject.toml
```diff
@@ -26,5 +26,5 @@ twine = "*"
 urllib3 = { version = "*", python = "<4"}
 
 [build-system]
-requires = ["poetry>=1.0.0"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core"]
```
Korijn, line 29: Doesn't this line normally include a minimum version?

berendkleinhaneveld: I created a new project with poetry (`poetry init`) and looked at the output. Apparently the number is not included anymore. Probably because poetry.core only has a 1.0.0 or bigger release?

Korijn (approved):

PR #93, opened 2023-10-25
Add support for Pyside6's new asyncio support

### LINKED ISSUE #91: PySide 6.6 asyncio support
Great news from the PySide camp: https://www.qt.io/blog/qt-for-python-6.6

An official asyncio compatible Qt event loop!

I guess it means we can replace Qt integration with standard asyncio logic and expect a performance increase.

Relevant code:

https://github.com/fork-tongue/observ/blob/9e016b74f5e4070cbc7db448e0e8020652e7ed11/observ/scheduler.py#L34-L53

"Standard asyncio logic" meaning something like this:

```python
     import asyncio 
  
     def request(deadline): 
         loop = asyncio.get_event_loop_policy().get_event_loop() 
         loop.call_soon(do_stuff, deadline) 
```

berendkleinhaneveld:
Fixes #91 .

### DIFF
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -5,6 +5,10 @@ and should be integrated in the event loop of your choosing.
 from bisect import bisect
 from collections import defaultdict
 import importlib
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 class Scheduler:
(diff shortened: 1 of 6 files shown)


observ/scheduler.py
```diff
@@ -44,13 +48,27 @@ def register_qt(self):
...
+
+            asyncio.set_event_loop_policy(QtAsyncio.QAsyncioEventLoopPolicy())
+
+            def request_flush():
+                loop = asyncio.get_event_loop_policy().get_event_loop()
+                loop.call_soon(scheduler.flush)
+
+            scheduler.register_request_flush(request_flush)
+        except ImportError:
+            logger.debug("Could not find QtAsyncio submodule, fall back to QTimer")
```
berendkleinhaneveld: Should I maybe turn this into a warning?

observ/scheduler.py
```diff
@@ -44,13 +48,27 @@ def register_qt(self):
...
-        self.timer.timeout.connect(scheduler.flush)
-        # Set interval to 0 to trigger the timer as soon
-        # as possible (when Qt is done processing events)
-        self.timer.setInterval(0)
-        self.register_request_flush(self.timer.start)
+        try:
+            QtAsyncio = importlib.import_module(f"{qt}.QtAsyncio")
+            import asyncio
+
+            asyncio.set_event_loop_policy(QtAsyncio.QAsyncioEventLoopPolicy())
```
Korijn: My expectation is that users of observ (application developers) will set the Qt event loop policy in their application initialization code.

Observ just needs to a `register_asyncio()` function that users can then call. It should be completely unaware of Qt. That makes observ compatible with any asyncio-compatible framework.

We can keep `register_qt()` around for backward compatibility.

Optionally, in `register_qt`,  you could try to detect if `QAsyncioEventLoopPolicy` is available issue a warning message to make users aware that they can upgrade.

berendkleinhaneveld: Makes a lot of sense!

berendkleinhaneveld: Fixed!

Korijn (approved):

PR #94, opened 2023-10-30
Store weak references to bound methods

berendkleinhaneveld:
This solves a problem where functions and callbacks store strong references to `self` which can prevent an object (and associated watchers) from being garbage collected.

- [x] convert weak decorator into a function that takes object and method as args

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -297,7 +297,9 @@ def weak(obj: Any, method: Callable):
 
         return wrapped
     else:
-        raise NotImplementedError
+        raise WrongNumberOfArgumentsError(
+            "Please use 1, 2 or 3 arguments for callbacks"
+        )
 
 
 def is_bound_method(fn: Callable):
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -260,16 +260,12 @@ class Watcher:
         return f"{self.fn.__module__}.{self.fn.__qualname__}"
 
 
-def weak(obj, method):
+def weak(obj: Any, method: Callable):
     """
-    Prevent the strong capture of the given object by using a weakref.ref instead.
-    Assumes the first argument to the method to be the argument for which a weak ref
-    is made.
-
-    The wrapped method will only be called, if the weak ref object is actually still
-    alive.
-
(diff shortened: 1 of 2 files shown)


Korijn (approved):

observ/watcher.py
```diff
@@ -317,3 +258,54 @@ def depend(self) -> None:
...
+    elif nr_arguments == 3:
+
+        @wraps(method)
+        def wrapped(new, old):
+            if this := weak_obj():
+                return method(this, new, old)
+
+        return wrapped
+    else:
+        raise NotImplementedError
```
Korijn: Doesn't this need brackets?

Korijn: Doesn't this need brackets?

berendkleinhaneveld: Actually, surprisingly, no, it doesn't :) But, your comment made me read the docs for NotImplementedError and I should not be using it here at all! A type error would be better suited.

berendkleinhaneveld:
Note that I've change the PySide example a bit: now the callback is a bound method from the QLabel class, but this poses all kinds of problems, which I tried to document in the test. I've marked it as xfail for now. If we ever encounter this in the real world, we might want to try and solve it, but I deemed it unnecessary for now.

tests/test_references.py
```diff
@@ -91,35 +95,61 @@ def cb():
...
+                # Use a method from QLabel directly as callback. These methods don't
+                # have a __func__ attribute, so currently this won't be wrapped by
+                # the 'weak' wrapper in the watcher, so it will create a strong
+                # reference instead to self.
+                # Ideally we would be able to detect this and wrap it, but then
+                # there is the problem that these kinds of methods don't have a
+                # signature, so we can't detect the number of arguments to supply.
+                # I guess we can assume that we should supply only two arguments
+                # in those cases: 'self' and 'new'?
+                # We'll solve this if this becomes an actual problem :)
```
Korijn, line 135: I guess you could specifically support Qt objects by checking if the object is an instance of QObject, but that's a big can of worms for another day...

PR #111, opened 2023-11-14
adds type hints to public APIs

### LINKED ISSUE #110: missing type hints for the public API
Hi,

The publis API does not have type hints ATM, which leads to a bad experience when using the library.

For example, the following expression:
```py
state = reactive({'value': 0})
```
is inferred as `Proxy | Any | tuple[Any, ...]`, so when trying to do this:
```py
state['value'] = 42
```
we get the following error: `"__getitem__" method not defined on type "Proxy"`.

Typescript [gets away with this](https://github.com/microsoft/TypeScript/blob/628bf0ec85be7c58019dc1742e50bc3f97168a5a/src/lib/es2015.proxy.d.ts#L108) by doing:
```ts
interface ProxyConstructor {
    new <T extends object>(target: T, ...): T;
}
```
so maybe we should have:
```py
def proxy[T](target: T, **etc) -> T: ...
```

ju1ius:
Closes #110

### DIFF
--- a/observ/proxy.py
+++ b/observ/proxy.py
@@ -1,9 +1,14 @@
+from __future__ import annotations
+
 from functools import partial
+from typing import cast, Generic, Literal, TypedDict, TypeVar
 
 from .proxy_db import proxy_db
 
+T = TypeVar("T")
 
-class Proxy:
+
+class Proxy(Generic[T]):
     """
     Proxy for an object/target.
 
@@ -16,9 +21,9 @@ class Proxy:
     """
 
     __hash__ = None
-    __slots__ = ["target", "readonly", "shallow", "__weakref__"]
+    __slots__ = ("target", "readonly", "shallow", "__weakref__")
 
-    def __init__(self, target, readonly=False, shallow=False):
+    def __init__(self, target: T, readonly=False, shallow=False):
         self.target = target
         self.readonly = readonly
         self.shallow = shallow
@@ -33,7 +38,7 @@ class Proxy:
 TYPE_LOOKUP = {}
 
 
-def proxy(target, readonly=False, shallow=False):
+def proxy(target: T, readonly=False, shallow=False) -> T:
     """
     Returns a Proxy for the given object. If a proxy for the given
     configuration already exists, it will return that instead of
@@ -67,14 +72,26 @@ def proxy(target, readonly=False, shallow=False):
             return proxy_type(target, readonly=readonly, shallow=shallow)
 
     if isinstance(target, tuple):
-        return tuple(proxy(x, readonly=readonly, shallow=shallow) for x in target)
+        return cast(
+            T, tuple(proxy(x, readonly=readonly, shallow=shallow) for x in target)
+        )
 
     # We can't proxy a plain value
-    return target
+    return cast(T, target)
 
 
-def ref(target):
-    return proxy({"value": target})
+try:
+    # for Python >= 3.11
+    class Ref(TypedDict, Generic[T]):
+        value: T
+
+    def ref(target: T) -> Ref[T]:
+        return proxy(Ref(value=target))
+
+except TypeError:
+    # before python 3.11 a TypedDict cannot inherit from a non-TypedDict class
+    def ref(target: T) -> dict[Literal["value"], T]:
+        return proxy({"value": target})
 
 
 reactive = proxy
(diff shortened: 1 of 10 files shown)


observ/proxy.py
```diff
@@ -33,7 +36,7 @@ def __del__(self):
 TYPE_LOOKUP = {}
 
 
-def proxy(target, readonly=False, shallow=False):
+def proxy(target: T, readonly=False, shallow=False) -> T:
```
Korijn, line 41: Shouldn't this be `-> Proxy[T]`?

observ/proxy.py
```diff
@@ -92,15 +101,15 @@ def to_raw(target):
...
-        return {key: to_raw(value) for key, value in target.items()}
+        return cast(T, {key: to_raw(value) for key, value in target.items()})
 
     if isinstance(target, tuple):
-        return tuple(to_raw(t) for t in target)
+        return cast(T, tuple(to_raw(t) for t in target))
 
     if isinstance(target, set):
-        return {to_raw(t) for t in target}
+        return cast(T, {to_raw(t) for t in target})
```
Korijn, line 121: are these `casts` in `to_raw` really necessary? `return {}` is already a plain dict right?

observ/watcher.py
```diff
@@ -9,27 +9,28 @@
...
 from .dict_proxy import DictProxyBase
 from .list_proxy import ListProxyBase
-from .proxy import Proxy
 from .scheduler import scheduler
 from .set_proxy import SetProxyBase
 
 
-T = TypeVar("T", bound=Callable[[], Any])
+T = TypeVar("T")
+Watchable = Callable[[], T] | T
```
Korijn: Feels like some detail was lost here, not everything is watchable. We support a callable, a proxy, or a list of proxies.

observ/proxy.py
```diff
@@ -16,9 +19,9 @@ class Proxy:
     """
 
     __hash__ = None
-    __slots__ = ["target", "readonly", "shallow", "__weakref__"]
+    __slots__ = ("target", "readonly", "shallow", "__weakref__")
```
Korijn, line 24: There's a few more places where we define `__slots__` with `[]`, you want to update those to `()` as well?

Korijn (changes requested):
Awesome work, thank you! I just finished reviewing, it looks really good and I learned a few new things. :)

observ/proxy.py
ju1ius: > Shouldn't this be -> Proxy[T]

The correct type would be `Proxy[T] | T`, since types like `int`, `float`, `str` or class instances are not proxied.

However for this type to be useful as a return type, you would need a way to express the fact that a `Proxy[T]` behaves _exactly_ like a `T` (i.e. delegates all methods to its inner `T`), which is not possible in the current state of the Python type system AFAICT.

observ/proxy.py
ju1ius: Ah no, that was not intended...
Passing a list to `__slots__` is useless since slots are supposed to be immutable, but I guess this change slipped through the formatter... Do you want me to revert this line?

observ/proxy.py
Korijn: I see, that's a shame. Either way, observ provides fully transparent wrappers over the builtin types, so it should be fine! :)

observ/proxy.py
Korijn: I like the change you made - switching to tuples! Please keep it and apply it to all instances of `__slots__` :)

Korijn:
Are you going to address the remaining comments? Otherwise I'd like to merge and move forward

ju1ius:
> Are you going to address the remaining comments? Otherwise I'd like to merge and move forward

If it's OK for you then by all means do! 👍🏻

Korijn (approved):

Korijn:
Thanks a lot @ju1ius <3

I've documented the two remaining comments in new issue #114

observ/watcher.py
Korijn: Not applicable, see discussion here: https://github.com/fork-tongue/observ/pull/112#discussion_r1397113343

observ/proxy.py
Korijn: @ju1ius can you share your thoughts on this one?

PR #112, opened 2023-11-14
Add async/await support for watch expressions and callbacks

### LINKED ISSUE #102: Asyncio support
Support `async` watch expressions and callbacks!

### LINKED ISSUE #115: Initialize method
After #112 is merged, I want to add an `init` function to the public API that will automatically configure:

* asyncio eager task factory
* register the scheduler with asyncio and/or Qt

All depending on the versions of libs available and if they are already configured or not.

The function should also have an `async` version so it is compatible with `asyncio.run`. It probably just calls the synchronous function.

Korijn:
Closes #102
Closes #115 

* `async def` callbacks
* `async def` watch expressions (most notably for `watch_effect`)

True async (meaning `sync=False`) watch expressions (not callbacks) are only supported starting with python 3.12 with [`asyncio.eager_task_factory`](https://docs.python.org/3/library/asyncio-task.html?highlight=eager_task_factory#asyncio.eager_task_factory)

### DIFF
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -79,7 +79,7 @@ class Scheduler:
             warnings.warn(
                 "QtAsyncio module available: please consider using `register_asyncio` "
                 "and call the following code:\n"
-                f"    from {qt} import QtAsyncio\n"
+                f"    from {qt} import QtAsyncio\n"  # noqa: E272
                 "    asyncio.set_event_loop_policy(QtAsyncio.QAsyncioEventLoopPolicy())"
                 ""
             )
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -2,6 +2,7 @@
 The scheduler queues up and deduplicates re-evaluation of lazy Watchers
 and should be integrated in the event loop of your choosing.
 """
+import asyncio
 from bisect import bisect
 from collections import defaultdict
 import importlib
@@ -44,17 +45,15 @@ class Scheduler:
         """
         self.request_flush = callback
 
+    def request_flush_asyncio(self):
+        loop = asyncio.get_event_loop_policy().get_event_loop()
+        loop.call_soon(self.flush)
+
     def register_asyncio(self):
         """
         Utility function for integration with asyncio
         """
-        import asyncio
-
-        def request_flush():
-            loop = asyncio.get_event_loop_policy().get_event_loop()
-            loop.call_soon(scheduler.flush)
-
-        scheduler.register_request_flush(request_flush)
+        self.register_request_flush(self.request_flush_asyncio)
 
     def register_qt(self):
         """
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -25,7 +25,6 @@ Watchable = Union[
     Callable[[], T],
     Callable[[], Awaitable[T]],
     T,
-    list[T],
 ]
 WatchCallback = Union[Callable[[], Any], Callable[[T], Any], Callable[[T, T], Any]]
 
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -236,7 +236,7 @@ class Watcher:
 
         if self.callback_async and maybe_coro:
             loop = asyncio.get_event_loop_policy().get_event_loop()
-            if self.sync and not loop.is_running():
+            if not loop.is_running():
                 loop.run_until_complete(maybe_coro)
             else:
                 loop.create_task(maybe_coro)
@@ -267,7 +267,7 @@ class Watcher:
             value_or_coro = self.fn()
             if self.fn_async and value_or_coro:
                 loop = asyncio.get_event_loop_policy().get_event_loop()
-                if self.sync and not loop.is_running():
+                if not loop.is_running():
                     loop.run_until_complete(value_or_coro)
                 else:
                     loop.create_task(value_or_coro)
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -154,18 +154,6 @@ class Watcher:
             else:
                 self.fn = fn
             self.fn_async = inspect.iscoroutinefunction(fn)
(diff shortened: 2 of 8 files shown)


Korijn:
There's a ton of differences between the way promises work in JS and in Python. For example, a promise will always execute eagerly to completion in JS, whereas in Python they will only complete if you explicitly await them. This makes the concept pretty difficult to implement... parking this for now.

Korijn:
I have overlooked the `weak` function which wraps functions. That needs to support coroutines too.

Korijn:
I think this is ready now!

tests/test_asyncio.py
```diff
@@ -0,0 +1,416 @@
...
+        assert (called, completed) == (1, 1)
+    flush(loop)
+    if create_loop is create_plain_loop and expr_async and create_async:
+        assert (called, completed) == (0, 0)
+    elif (
+        create_loop is create_eager_loop and expr_async and create_async and write_async
+    ):
+        assert (called, completed) == (0, 0)
+    else:
+        assert (called, completed) == (1, 1)
```
Korijn: This section of the test is a little too much to handle for my brains. I just reverse engineered the `assert` statements from the outputs. I _think_ they make sense... but it's a lot to process.

ju1ius: IMO you should write one test function for each combination:
```py
def test_sync_watchable_with_sync_callback(): ...
def test_sync_watchable_with_async_callback(): ...

def test_async_watchable_with_sync_callback(): ...
def test_async_watchable_with_async_callback(): ...
```
Then maybe move the setup boilerplate to helper functions as needed.

At some point, and _especially_ in tests, some degree of duplication is better than combinatorial logic explosion that is hard to follow and refactor.

ju1ius (changes requested):
@Korijn You should also update the `Watchable` type:

```py
Watchable = Union[
    Callable[[], T],
    Callable[[], Coroutine[Any, Any, T]],
    T,
]
```

tests/test_asyncio.py
Korijn: Well, yes, but there are 48 cases in total... :(

In this case, it's pytest generating the cases. I can debug each individual case and check if it's acting correctly. I just need to go through all the cases one by one here and see if the applicable asserts make sense.

For `test_asyncio_watcheffect_multi`, the asserts are quite easy to read and they make sense to me. It confirms the implementation is correct.

For `test_asyncio_watcher_multi`, it's more complex. I just need a bit more time to dig in and convince myself it's correct.

Korijn: After reviewing more carefully, I think only the case `create_loop is create_eager_loop and expr_async and create_async and write_async` is suspicious and might be a bug. I'll step through that one.

Korijn: I figured it out, no bugs! I documented all the assert cases with extensive code comments.

Korijn:
OK, for real this time, @berendkleinhaneveld, can you review this?

observ/watcher.py
```diff
@@ -20,7 +21,12 @@
 
 
 T = TypeVar("T")
-Watchable = Union[Callable[[], T], T]
+Watchable = Union[
+    Callable[[], T],
+    Callable[[], Awaitable[T]],
+    T,
+    list[T],
```
ju1ius: I'm not sure what you're trying to achieve here?

ju1ius: This:
```py
a = proxy({'value': 'a'})
b = proxy({'value': 42})
w = watch([a, b], print, immediate=True, sync=True)
```
Should print
```py
[{'value': 'a'}, {'value': 42}]
```
Right?

If so, then you must _not_ include `list[T]` in the union.

ju1ius: Let's test this:

```py
Watchable = Union[Callable[[], T], T]

a = proxy({'value': 'a'})
b = proxy({'value': 42})
w = watch([a, b], print, immediate=True, sync=True)
# [{'value': 'a'}, {'value': 42}]
typing.reveal_type(w) # Type of "w" is Watcher[list[dict[str, str] | dict[str: int]]]
```
Which is what we expect, but
```py
Watchable = Union[Callable[[], T], T, list[T]]

a = proxy({'value': 'a'})
b = proxy({'value': 42})
w = watch([a, b], print, immediate=True, sync=True)
# [{'value': 'a'}, {'value': 42}]
typing.reveal_type(w) # Type of "w" is Watcher[dict[str, str] | dict[str: int]]
```
is not.

ju1ius:
```suggestion
```

ju1ius (changes requested):

observ/watcher.py
Korijn: You're right, of course! That makes sense.

I really appreciate all the explanation and code review you're doing, I'm learning a lot from you. I'm not very experienced with type hints. Thanks!

ju1ius (approved):

observ/scheduler.py
```diff
@@ -79,7 +79,7 @@ def register_qt(self):
             warnings.warn(
                 "QtAsyncio module available: please consider using `register_asyncio` "
                 "and call the following code:\n"
-                f"    from {qt} import QtAsyncio\n"
+                f"    from {qt} import QtAsyncio\n"  # noqa: E272
```
berendkleinhaneveld, line 81: Was this put here intentionally?

berendkleinhaneveld (approved):
I have read most of the comments, but its a bit much for me, parsing them and really understanding them within a limited timeframe. I say: LGTM: I think it looks great! And I think it is in a great state should we encounter problems later on and have to dig in again.

observ/scheduler.py
Korijn: Yes. It only triggers when linting on python 3.12 :/ I ran into that while working locally

PR #113, opened 2023-11-14
Add global Watcher lifecycle hooks

### LINKED ISSUE #107: Provide global Watcher lifecycle hooks
At least created and destroy.

```py
off = Watcher.on("created", hook)
off = Watcher.on("destroy", hook)
```

Korijn:
Closes #107

This makes it possible to tie the Watcher lifecycle to something global instead of local.

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -129,8 +129,8 @@ class Watcher:
         "_number_of_callback_args",
         "__weakref__",
     ]
-    on_created = None
-    on_destroyed = None
+    on_created: Optional[Callable[[Watcher], None]] = None
+    on_destroyed: Optional[Callable[[Watcher], None]] = None
 
     def __init__(
         self,
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -129,6 +129,8 @@ class Watcher:
         "_number_of_callback_args",
         "__weakref__",
     ]
+    on_created = None
+    on_destroyed = None
 
     def __init__(
         self,
@@ -140,7 +142,7 @@ class Watcher:
     ) -> None:
         """
         sync: Ignore the scheduler
-        lazy: Only reevalutate when value is requested
+        lazy: Only reevaluate when value is requested
         deep: Deep watch the watched value
         callback: Method to call when value has changed
         """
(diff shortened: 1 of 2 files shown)


observ/watcher.py
```diff
@@ -129,6 +129,8 @@ class Watcher:
         "_number_of_callback_args",
         "__weakref__",
     ]
+    on_created = None
```
berendkleinhaneveld: This would benefit from type annotations, so maybe wait for #111 ? Or otherwise we'll need to update it in that PR.

tests/test_watcher_hooks.py
```diff
@@ -0,0 +1,35 @@
...
+
+        def _callback():
+            nonlocal called
+            called += 1
+
+        watch(lambda: len(a), _callback, sync=True)
+        assert called == 0
+        a.append(3)
+        assert called == 1
+        assert len(a) == 3
```
berendkleinhaneveld, line 32: I'm missing a check for the length of the watchers array here. In its current state it is not testing that much :)

berendkleinhaneveld (changes requested):

tests/test_watcher_hooks.py
Korijn: Well without the watchers array, it won't be called because the watcher is gc'd!

Korijn: I extended the test anyway :)

observ/watcher.py
Korijn: Added type hints

observ/watcher.py
```diff
@@ -130,6 +130,8 @@ class Watcher(Generic[T]):
         "_number_of_callback_args",
         "__weakref__",
     )
+    on_created: Optional[Callable[[Watcher], None]] = None
+    on_destroyed: Optional[Callable[[Watcher], None]] = None
```
ju1ius:
```suggestion
    on_created: Optional[Callable[[Watcher[T]], None]] = None
    on_destroyed: Optional[Callable[[Watcher[T]], None]] = None
```

ju1ius (changes requested):

observ/watcher.py
```diff
@@ -137,6 +137,8 @@ class Watcher(Generic[T]):
         "_number_of_callback_args",
         "__weakref__",
     )
+    on_created: Optional[Callable[[Watcher[T]], None]] = None
```
berendkleinhaneveld, line 140: Because we use `from __future__ import annotations` we can actually use:
`Callable[[Watcher[T]], None] | None`. And also replace `Union` with `|`. 
It just came to mind, and it's out of scope for this PR. I'll open a separate issue.

berendkleinhaneveld (approved):

observ/watcher.py
berendkleinhaneveld: I've actually updated #114  with this idea.

PR #118, opened 2023-11-28
Replace flake8 and black with ruff

berendkleinhaneveld:
Just noticed that ruff has a 0.1.* version out that includes a formatter that serves as a drop-in replacement for black.
Thought I'd try to apply it to this repo, see what it would entail to adopt it.

One thing I noticed is that the linter will not complain if something is wrong that could be solved by running the formatter, so in the pre-commit config and in the linting step in CI, I've added a step to check the formatting specifically.


pyproject.toml
```diff
@@ -12,27 +15,45 @@ python = ">=3.9"
...
+    "E4",
+    "E7",
+    "E9",
+    "F",   # Pyflakes (default)
+    "I",   # isort imports
+    "N",   # pep8-naming
+    "T10", # flake8-debugger
+    "T20", # flake8-print
+    "RUF", # ruff
+]
```
Korijn, line 47: Can we also enable [RUF100](https://docs.astral.sh/ruff/linter/#detecting-unused-suppression-comments)? It looks really useful.

Korijn: Nevermind, it's already enabled.

.pre-commit-config.yaml
```diff
@@ -1,16 +1,20 @@
 repos:
 - repo: local
   hooks:
-  - id: flake8
-    name: Linting
-    entry: poetry run flake8
+  - id: ruff (format)
+    name: Format
+    entry: poetry run ruff format --check
```
Korijn: are you sure about using --check here?

pyproject.toml
```diff
@@ -12,27 +15,45 @@ python = ">=3.9"
...
-PySide6 = { version = ">=6.6", python = "<3.13"}
+PySide6 = { version = ">=6.6", python = "<3.13" }
 pytest-qt = "*"
 pytest-xvfb = "*"
 
+[tool.ruff.lint]
+select = [
+    "E4",
+    "E7",
+    "E9",
```
Korijn, line 40: I see you left out E5 (line too long) which I agree is super annoying. But does the formatter still apply max line length? I think we still want to maintain a max line length, right?

.pre-commit-config.yaml
berendkleinhaneveld: Well, depends whether you would want this step to automatically format the code for you. Which could be nice, but it seemed maybe better to start simple? But yeah, I'm actually all for removing the check part :)

pyproject.toml
berendkleinhaneveld: I picked the default set of rules from ruff and expanded on it a bit. E5 just wasn't in there.

> While the [line-too-long](https://docs.astral.sh/rules/line-too-long/) (E501) rule can be used alongside the formatter, the formatter only makes a best-effort attempt to wrap lines at the configured [line-length](https://docs.astral.sh/settings/#line-length). As such, formatted code may exceed the line length, leading to [line-too-long](https://docs.astral.sh/rules/line-too-long/) (E501) errors. 
(https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules)

So the formatter tries its best to keep line-length in check (default is 88, like black).

On the one hand, it might be annoying to have E501, but long lines themselves can be annoying as well... So I'd rather have shorter lines :)

Korijn (approved):

berendkleinhaneveld:
I got one more thing: I also wanted to checkout the default setting for import sorting, so I added a commit that does just that. I'm not sure I like the order much better, but on the other hand: as long as I don't have to order the imports by hand, I'm happy. What do you think?

And I also applied `--fix` on the linting step in pre-commit, so it mirrors the behaviour of the format step.

Korijn:
> I got one more thing: I also wanted to checkout the default setting for import sorting, so I added a commit that does just that. I'm not sure I like the order much better, but on the other hand: as long as I don't have to order the imports by hand, I'm happy. What do you think?
> 
> And I also applied `--fix` on the linting step in pre-commit, so it mirrors the behaviour of the format step.

The imports are still in three groups right? That's really what's most important to me. I want to be able to quickly see what standard library and third party dependencies are used. I don't care about the order within the groups.

Seems fine with me.

PR #120, opened 2024-01-02
Object proxies

### LINKED ISSUE #103: Support for dataclasses and / or objects with simple attributes
We might want to consider adding support for dataclasses and/or attributes of objects. Or maybe provide an API to make certain attributes 'reactive' for custom classes/objects. 

If we decide against it, this can be a trace of that decision :)

Korijn:
Closes #103

🚧🚧🚧 Work in progress

I'm just starting to get to grips with the complexity of this topic. It's been a long time since I looked at object proxying.

The boilerplate is done.

The main complexity will be in `object_proxy.py`. In particular the challenge is that there's no watertight method of trapping attribute access only, without also trapping method calls. So for example

```py
class Foo:
    def __init__(self):
        self.bar = 5

    def baz(self):
        return 5

    def quux(self):
        return self.bar

f = Foo()
```

* It's obvious that `f.bar` should be reactive.
* However I think `f.quux` should not be reactive (since `f.bar` is already reactive).
* `f.baz` is not stateful, so it doesn't need to be reactive.

The list of attributes to trap is dynamic, like the keys of a dict, but we cannot replace the internal `f.__dict__`. All we have to work with is a trap on `f.__getattribute__` but that also fires for the `f.bar` and `f.quux`.

Another challenge is that on objects, you can define something like `__iter__` to make a type iterable, while by default classes don't have an `__iter__` implementation.

🤔🤔🤔💭💭💭 Needs a lot more thought...

### DIFF
--- a/observ/dict_proxy.py
+++ b/observ/dict_proxy.py
@@ -75,4 +75,9 @@ ReadonlyDictProxy = type(
     },
 )
 
-TYPE_LOOKUP[dict] = (DictProxy, ReadonlyDictProxy)
+
+def type_test(target):
+    return isinstance(target, dict)
+
+
+TYPE_LOOKUP[type_test] = (DictProxy, ReadonlyDictProxy)
--- a/observ/list_proxy.py
+++ b/observ/list_proxy.py
@@ -69,4 +69,8 @@ ReadonlyListProxy = type(
 )
 
 
-TYPE_LOOKUP[list] = (ListProxy, ReadonlyListProxy)
+def type_test(target):
+    return isinstance(target, list)
+
+
+TYPE_LOOKUP[type_test] = (ListProxy, ReadonlyListProxy)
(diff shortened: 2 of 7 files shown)


Korijn:
I've spent another two hours on this and am abandoning this feature (again). It just isn't possible in a generic sense. Good luck to the next guy. Problems in a nutshell:

* proxying all attribute gets and sets generically is very complicated
  * proxying `__getattribute__` and `__setattr__` gets you a long way, however instance methods, static methods, class methods, instance attributes and class attributes all flow through these functions so you need a mechanism to distinguish state from logic and I couldn't find one that's foolproof
  * if you proxy `__getattribute__` on `Proxy`, that means `Proxy.__init__` will also trigger the traps but the traps rely on `Proxy.__init__` to have already completed
  * magic methods must be proxied since they bypass `__getattribute__`
  * most magic methods (e.g. `__len__`) are optional on classes
* unlike dict, list and set, there is no generic `.copy()` method you can rely on so detecting if something has actually changed is difficult
* how do you traverse a class? which attributes should you traverse? how do you exclude the attributes defined on `Proxy` (such as `target`)?

... and a lot more problems

The next best alternative that _does_ work in a generic sense is to declare a reactive dict on your class and work with that.

PR #121, opened 2024-01-05
Object proxy

### LINKED ISSUE #103: Support for dataclasses and / or objects with simple attributes

Korijn:
Closes #103

I think I got it down pretty nicely this time.

It required a very different approach, and a way to distinguish stateful attributes from callables.


observ/object_proxy.py
```diff
@@ -0,0 +1,254 @@
...
+                # so no need to track this modification
+                # TODO: we could cache this
+                return retval
+
+        new_value = getattr(target, name, None)
+        if is_new_keydep:
+            attrs["keydep"][name] = Dep()
+        if xor(old_value is None, new_value is None) or old_value != new_value:
+            attrs["keydep"][name].notify()
+            attrs["dep"].notify()
```
Korijn: I think this dep is unnecessary actually? There's no way to depend on the object as a whole, I think... 🤔

observ/object_utils.py
```diff
@@ -0,0 +1,24 @@
...
+from itertools import chain
+
+
+@cache
+def get_class_slots(cls):
+    """utility to collect all __slots__ entries for a given type and its supertypes"""
+    # collect via iterables for performance
+    # deduplicate via set
+    return set(
+        chain.from_iterable(getattr(cls, "__slots__", []) for cls in cls.__mro__)
```
berendkleinhaneveld, line 11: Cool trick!

observ/proxy.py
```diff
@@ -21,12 +21,14 @@ class Proxy(Generic[T]):
     """
 
     __hash__ = None
-    __slots__ = ("target", "readonly", "shallow", "__weakref__")
+    # the slots have to be very unique since we also proxy objects
+    # which may define the attributes with the same names
+    __slots__ = ("__target__", "__readonly__", "__shallow__", "__weakref__")
```
berendkleinhaneveld, line 26: Since they need to be very unique, did you consider using a prefix such as `__observ_`?

observ/watcher.py
```diff
@@ -95,11 +103,7 @@ def traverse(obj, seen=None):
         seen = []
     seen.append(obj)
     for v in val_iter:
-        if (
-            isinstance(
-                v, (dict, DictProxyBase, list, ListProxyBase, set, SetProxyBase, tuple)
-            )
-        ) and v not in seen:
+        if v not in seen:
```
berendkleinhaneveld, line 107: Wasn't there a reason that we checked the type here? Now that I'm thinking about it: if it would have been important I think we would have added a regression test, I hope 😅

observ/object_proxy.py
```diff
@@ -0,0 +1,254 @@
...
+            # should be reactive, so let's determine it now
+            target_attrs_new = get_object_attrs(target)
+            if name not in target_attrs_new:
+                # the set attr is not stateful (e.g. someone
+                # is attaching a bound method)
+                # so no need to track this modification
+                # TODO: we could cache this
+                return retval
+
+        new_value = getattr(target, name, None)
```
berendkleinhaneveld, line 56: Why is this `getattr` here? Isn't the `retval` the same as `new_value`? Maybe I'm missing some kind of nuance.

observ/object_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+
+
+def get_object_attrs(obj):
+    """utility to collect all stateful attributes of an object"""
+    # __slots__ from full class ancestry
+    attrs = list(get_class_slots(type(obj)))
+    try:
+        # all __dict__ entries
+        attrs.extend(vars(obj).keys())
```
berendkleinhaneveld: Is there a need to return a list instead of a set? Wouldn't it be better and maybe more performant (on the call site) to return a set instead?

berendkleinhaneveld (commented):
Nice surprise to see that you've managed to tackle this! 

Would love to play around with this. I guess the test suite might need some more love, but that can also wait until we've tried it out for a bit.

Would be nice to see an update to the benchmark to get a sense of how it affects performance.

Just came to mind: does the `to_raw` method still work? Or did I miss that in the review?

observ/proxy.py
Korijn: Yeah I did and it was really annoying to read... We can still do it.

observ/watcher.py
Korijn: Well the thing is I have to add `object` to the list now, and everything in python is an object, so the check becomes useless.

Korijn:
> Just came to mind: does the `to_raw` method still work? Or did I miss that in the review?

Not for objects. I can't think of a way to implement it.

observ/object_utils.py
Korijn: `get_object_attrs` is really the silver bullet of this PR that makes it all possible

observ/object_proxy.py
Korijn: `setattr` doesn't return anything I think. We could do without `retval` perhaps.

In fact we can just use `value` directly...

berendkleinhaneveld: How would this work with property getters/setters that change the inputs? Maybe we actually do need the getattr.

observ/proxy.py
berendkleinhaneveld: I can imagine that :) Let's keep it like it is now.

observ/object_proxy.py
Korijn: Yeah.. That occurred to me as well

Korijn:
> Nice surprise to see that you've managed to tackle this!

I am very relieved that I can finally put this to rest. It's been in the back of my mind for years. 🤣 

I have to say I am surprised this works as well as it does. It's still technically possible to circumvent the mechanisms here, but it requires very specific, outlandish setups.

For example, ndarrays are objects but I do not think they will be reactive. We have to test I guess.

observ/object_proxy.py
Korijn: Removed it

observ/object_proxy.py
Korijn: Reviewed the code another time, doesn't need to change

observ/object_utils.py
Korijn: In order to avoid mutating the cached `get_class_slots` return value we have to create a copy at least. I thought a list would be faster. But it's probably negligible and there could still be duplicates between `__slots__` and `__dict__` potentially? Anyway I've changed it to avoid the copy if `__dict__` is empty or missing.

Korijn:
> For example, ndarrays are objects but I do not think they will be reactive. We have to test I guess.

ndarrays (and memoryviews) are definitely not reactive. We would have to write a separate `ndarray_proxy.py` module for that and I would not recommend that anybody use it. :) So I extended the `type_test` to exclude ndarray's from proxying.

observ/object_proxy.py
```diff
@@ -0,0 +1,255 @@
...
+            attrs["keydep"][name].notify()
+            del attrs["keydep"][name]
+
+        return retval
+
+
+def passthrough(method):
+    def trap(self, *args, **kwargs):
+        fn = getattr(self.__target__, method, None)
+        if fn is None:
```
Korijn, line 88: it's possible an attribute exists and is None, raising a TypeError

berendkleinhaneveld: Add it in a test please :)

Korijn: Done. Turns out it was already working properly.

berendkleinhaneveld (approved):

PR #130, opened 2025-09-02
Fix compatibility with pygfx

berendkleinhaneveld:
This change prevents reactivity on custom dict subclasses and instead will check that only 'real' (or pure) dicts are converted into reactive proxies.
The problem is that pygfx's geometry is a dict subclass and turning it into a reactive proxy will stop it from being recognized as a Geometry type, so pygfx will not accept it when you try to set the geometry attribute on a Mesh.
I could check for pygfx specifically, but it seems to me that more problems are to be expected in the future if observ tries to proxy custom dict subclasses.
I'm thinking to update list and set proxy as well.

A workaround in Collagraph is to check for the `__target__` attribute when setting the attribute, but I guess that this change also prevents any computational overhead from creating unnecessary proxies of dict subclasses.


uv.lock
Korijn, line 1: Maybe not check in the uv lockfile?

berendkleinhaneveld: Sure, just checked to see that `uv sync` still works without lock file. Will update .gitignore

PR #136, opened 2025-11-14
Separate benchmark workflow

berendkleinhaneveld:
Run benchmark as separate workflow and compare benchmark on PRs with latest from master (if available). Uses the cache action to store the results.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,4 +1,4 @@
-name: Performance Benchmarks
+name: Benchmarks
 
 on:
   push:
@@ -10,7 +10,7 @@ on:
 
 jobs:
   benchmark:
-    name: Run performance benchmarks
+    name: Benchmarks
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,78 @@
+name: Performance Benchmarks
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  benchmark:
+    name: Run performance benchmarks
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,78 @@
...
+      - master
+
+jobs:
+  benchmark:
+    name: Run performance benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0  # Need full history for comparisons
```
Korijn: Really? (Looks like AI nonsense?)

berendkleinhaneveld: Yes, indeed, let me put this to draft :D

berendkleinhaneveld: Ok, sorry about that! Should be a bit better now! Back to sleep now :')

.github/workflows/benchmark.yml
```diff
@@ -26,17 +26,16 @@ jobs:
...
+      - name: Restore benchmark baseline
+        uses: actions/cache/restore@v4
         with:
           path: .benchmarks
-          key: benchmark-results-${{ runner.os }}-${{ github.ref_name }}
+          key: benchmark-baseline-${{ runner.os }}-${{ github.sha }}
           restore-keys: |
-            benchmark-results-${{ runner.os }}-master
-            benchmark-results-${{ runner.os }}-
+            benchmark-baseline-${{ runner.os }}-
```
Korijn: Does this ever match anything?

berendkleinhaneveld: Yes, after the first run on master it will. But, now that I think about it a bit more: the -{{github.sha}} part is pretty useless. I'm only interested in a cache that depends on the runner, and maybe on the Python version used. So I can just make a specific key for that without the sha.

berendkleinhaneveld: Fixed!

Korijn (approved):

PR #138, opened 2025-11-27
Disable object reactivity

berendkleinhaneveld:
Disable object reactivity since it is slowly becoming a can of worms.


berendkleinhaneveld:
Context: I've run into another problem with `pygfx` + `observ`:
```
  File "./.venv/lib/python3.13/site-packages/observ/watcher.py", line 116, in traverse
    for v in val_iter:
             ^^^^^^^^
  File "./.venv/lib/python3.13/site-packages/observ/watcher.py", line 105, in <genexpr>
    val_iter = (getattr(obj, attr) for attr in obj_attrs)
                ~~~~~~~^^^^^^^^^^^
AttributeError: 'pygfx.utils.transform.RecursiveTransform' object has no attribute 'cache__directions_cache'. Did you mean: 'cache__decomposed_cache'?
```

I could also investigate a bit more into what happens here, but I think `pygfx` is a lib that actually does some clever tricks which make interoperability with `observ` much harder. In order to be more predictable, maybe it is a better idea to disable the object reactivity.

@Korijn What do you think?

Korijn:
I'm happy you finally came to your senses :)

(Your originally wanted this functionality right?)

berendkleinhaneveld:
I'm not really sure whether I really wanted this, but it was surely worth trying it out! Learning things (the hard way)!
Anyways: I'll do some more cleanup :)

Korijn:
Yes! Less code!

PR #142, opened 2025-12-12
Optimize traverse method

berendkleinhaneveld:
Try to optimize the traverse method by using a set with object ids which should yield faster lookups. However, this time also keep all the encountered objects referenced in a list so that for the duration of the method the object ids will not be re-used. See #71 for history on this.

In order to compare the performance to the master branch, I adjusted the benchmark workflow to run the benchmarks from the PR also with the code from master. This way the cache is not needed and any new benchmarks can still be compared with the code from master.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -25,7 +25,6 @@ jobs:
 
       # On PRs: run benchmarks twice (PR code vs master code) and compare
       - name: Run benchmarks on master code (baseline)
-        continue-on-error: true
         run: |
           # Checkout master version of observ directory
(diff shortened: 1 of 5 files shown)


Korijn:
If I can give any advice it is to wrap for loops in if statements and to use the walrus operator to create locals. This was the key for micro optimization in pygfx transforms.

berendkleinhaneveld:
image

Weird stuff going on in CI: difference in reported processor speed in between steps 😕

berendkleinhaneveld:
But I do like this part:
image

Korijn:
Love it!

PR #143, opened 2025-12-12
Convert recursive method into iterative to prevent function calls

berendkleinhaneveld:
Just after merging #142 I thought of making this method non-recursive to prevent the overhead of the recursive function call. This should improve the performance a bit more, although not as dramatically as #142.

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+from collections import deque
 from collections.abc import Awaitable, Container
 from functools import partial, wraps
 from itertools import count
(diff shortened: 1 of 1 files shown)


berendkleinhaneveld:
```
------------------------ benchmark 'traverse_deep': 2 tests ------------------------
Name (time in us)                          Mean             StdDev            Rounds
------------------------------------------------------------------------------------
test_traverse_deep (NOW)               899.7457 (1.0)      27.3784 (1.0)         712
test_traverse_deep (0001_master)     1,161.3173 (1.29)     59.9269 (2.19)        593
------------------------------------------------------------------------------------
```
Not too bad!

PR #148, opened 2025-12-15
Support stopping a watcher by calling it

berendkleinhaneveld:
This can help with disabling watchers at critical moments, for example right before they are removed. This alleviates the need for work-arounds such as installing a dummy fn property and clearing the callback property.

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -137,11 +137,11 @@ class WrongNumberOfArgumentsError(TypeError):
 class Watcher(Generic[T]):
     __slots__ = (
         "__weakref__",
+        "_active",
         "_deps",
         "_new_deps",
         "_number_of_callback_args",
         "_tasks",
-        "active",
         "callback",
         "callback_async",
         "deep",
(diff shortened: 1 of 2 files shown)


Korijn:
Vue also has this kind of option right?

berendkleinhaneveld:
Kind of: you can stop it with returned value: https://vuejs.org/guide/essentials/watchers.html#stopping-a-watcher
Instead, we return the watcher itself of course, so that the caller can manage its lifetime. But yeah, there is indeed something to make it stop explicitly if really needed

berendkleinhaneveld:
I could rename it 'stop' and make it a method instead that also unsets the callback and fn, so that resources could be released asap? That would make it more explicit and clear that the watcher has become useless after calling it.

Korijn:
If it still works for you I would prefer to align with vue's design when possible, otherwise feel free to diverge

berendkleinhaneveld:
I've now made it as in Vue, so you can call the watcher to stop it and clean up any resources.

Korijn:
Looks great!

PR #151, opened 2025-12-23
Figure out right amount of arguments for a callback without inspect.trace

Korijn:
Instead of letting `inspect` reconstruct the full trace, just use what's already available in the exception.

### DIFF
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -316,13 +316,14 @@ class Watcher(Generic[T]):
         try:
             return self.callback(*args)
         except TypeError as e:
-            frames = inspect.trace()
-            try:
-                if len(frames) != 1:
-                    raise
+            _run_callback_is_top_frame = (
+                e.__traceback__.tb_frame.f_code == self._run_callback.__code__
+            )
+            _no_lower_frames = e.__traceback__.tb_next is None
+            if _run_callback_is_top_frame and _no_lower_frames:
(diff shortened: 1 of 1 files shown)


observ/watcher.py
```diff
@@ -316,13 +316,25 @@ def _run_callback(self, *args) -> None:
...
-                if len(frames) != 1:
-                    raise
+                _run_callback_is_top_frame = (
+                    e.__traceback__.tb_frame.f_code == self._run_callback.__code__
+                )
+                _no_lower_frames = e.__traceback__.tb_next is None
+                wrong_number_of_arguments = (
+                    _run_callback_is_top_frame and _no_lower_frames
+                )
+            except AttributeError:
```
berendkleinhaneveld, line 330: Isn't it better to check for the existence of the attribute instead of doing a try/except?

Korijn: Actually no! This is less instructions. Try/except is very cheap in Python.

berendkleinhaneveld (approved):

observ/watcher.py
berendkleinhaneveld: I guess checking for a single attribute is probably faster without try/except, but there are multiple places where an AttributeError might be raised since some internal machinery is called, so I'm ok with this!

edit: in most cases the attribute access should work ao it is also a matter of how often you expect to hit the except. I'll refrain from saying anything about performance until I benchmarked it in a proper context 😅

PR #152, opened 2026-01-02
Optimize GC callback

Korijn:
This PR changes a few things:

* Adds early exits for GC collection generations 0 and 1: especially gen 0 trigger almost constantly and it benefits runtime greatly if it can do all that work without having to execute Python callbacks like this one
* Micro-optimized the rest of the logic so it has the absolute minimum amount of Python steps to perform - as most of the time when this callback runs it will have nothing to do, we want to find out as quickly as we can

Note - the benefits are pretty tricky to see in a benchmark. It's more noticeable in long running usage.

### DIFF
--- a/observ/proxy_db.py
+++ b/observ/proxy_db.py
@@ -27,21 +27,47 @@ class ProxyDb:
         Callback for garbage collector to cleanup the db for targets
         that have no other references outside of the db
         """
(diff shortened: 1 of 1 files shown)


berendkleinhaneveld (approved):
Nice!

PR #153, opened 2026-01-07
Fix for deletion of keys

### LINKED ISSUE #134: Calling pop with a default arg results in KeyError
```
  ...
  File "./file.cgx", line 66, in <lambda>
  File "./file.cgx", line 53, in local_props
    result.pop("children", None)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.13/site-packages/observ/traps.py", line 153, in trap
    attrs["keydep"][key].notify()
    ~~~~~~~~~~~~~~~^^^^^
KeyError: 'children'
```

berendkleinhaneveld:
Fixes #134 

There was a problem with `dict.pop(key, default)`, which threw a KeyError from the delete_key_trap. In order to fix this, it now checks before notifying whether the key actually existed in the dict before trying to delete the key to decide if it should notify any deps. Because if the key wasn't there in the first place, then there is no need to notify.

While investigating, I found another problem with the `.get(key)` method that after an existing key has been deleted, then any watcher using `.get(key)` will not be notified anymore, because the `keydep` had been removed. Now it includes a check to see if there are still any subscribers for this key, because if so: then the keydep should not be removed yet.

### DIFF
--- a/observ/traps.py
+++ b/observ/traps.py
@@ -146,12 +146,18 @@ def delete_key_trap(method, obj_cls):
 
     @wraps(fn)
     def trap(self, *args, **kwargs):
-        retval = fn(self.__target__, *args, **kwargs)
         key = args[0]
         attrs = proxy_db.attrs(self)
-        attrs["dep"].notify()
-        attrs["keydep"][key].notify()
-        del attrs["keydep"][key]
+        key_existed = key in self.__target__
+        retval = fn(self.__target__, *args, **kwargs)
+        if key_existed:
+            attrs["dep"].notify()
+            if key in attrs["keydep"]:
+                keydep = attrs["keydep"][key]
+                keydep.notify()
+                # Only delete keydep if no subscribers are interested in this key
+                if not keydep._subs:
+                    del attrs["keydep"][key]
         return retval
 
     return trap
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -359,4 +359,26 @@ def test_dict_delete_keynotify():
         getattr(coll, name)(*args[name])
         mocks[None].update.assert_called_once()
         mocks[key].update.assert_called_once()
-        assert len(proxy_db.attrs(coll)["keydep"]) == 0
+        # keydep entries are preserved so watchers can be notified when the key is re-added
+        assert key in proxy_db.attrs(coll)["keydep"]
+
+
+def test_dict_pop_with_default():
+    """Test that dict.pop with a default value works for missing keys."""
+    coll = DictProxy({2: 3})
+    mock = Mock()
+    proxy_db.attrs(coll)["dep"].add_sub(mock)
+
+    # Pop existing key - should notify and return value
+    result = coll.pop(2, "default")
+    assert result == 3
+    mock.update.assert_called_once()
+    # keydep is cleaned up when there are no subscribers
+    assert 2 not in proxy_db.attrs(coll)["keydep"]
+
+    # Pop missing key with default - should return default without error
+    mock.reset_mock()
+    result = coll.pop(999, "default")
+    assert result == "default"
+    # Should not notify since nothing changed
+    mock.update.assert_not_called()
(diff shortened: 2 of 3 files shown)


observ/traps.py
```diff
@@ -146,12 +146,18 @@ def delete_key_trap(method, obj_cls):
...
-        del attrs["keydep"][key]
+        key_existed = key in self.__target__
+        retval = fn(self.__target__, *args, **kwargs)
+        if key_existed:
+            attrs["dep"].notify()
+            if key in attrs["keydep"]:
+                keydep = attrs["keydep"][key]
+                keydep.notify()
+                # Only delete keydep if no subscribers are interested in this key
+                if not keydep._subs:
```
Korijn, line 159: why though? if the key is reintroduced later on, a new keydep will be created.

watchers should be triggered because of the main dep on the dict, and can then resubscribe to the new keydep.

right?

tests/test_collections.py
```diff
@@ -359,4 +359,27 @@ def new_mock(key=None):
         getattr(coll, name)(*args[name])
         mocks[None].update.assert_called_once()
         mocks[key].update.assert_called_once()
-        assert len(proxy_db.attrs(coll)["keydep"]) == 0
+        # keydep entries are preserved so watchers
+        # can be notified when the key is re-added
```
Korijn: Any changes in the set of keys should trigger the main dep which notifies watchers. All key watchers should also watch the main dep for this reason.

Korijn:
I'm not convinced this is right! Dicts have a main dep that triggers whenever the set of keys change (added or removed keys). That should trigger the watchers when the key is re-added.

berendkleinhaneveld:
I've extracted the test for re-adding the key here: #154 so you can see it fail :)

I'm trying to wrap my head around what is happening, but I think that the keydeps are not to be confused with the existence of keys in the actual target. The keydeps are populated on basis of the subscribers, not by whatever key is in the proxied object. So when a key is deleted, that should _not_ simply remove the deps for that key.

observ/traps.py
berendkleinhaneveld: There is no main dep on the dict. We could consider adding that and that will fix this issue (I checked that this afternoon), but that would be at the cost of way more false triggers of the watchers.

berendkleinhaneveld:
The more I think about it, the more this actually makes sense to me. As long as a watcher has a dependency on a key, then it should keep that dependency. No need to remove it. When the watcher is deleted, it will be removed from the `_subs` (WeakSet), so the subs will clear and are linked to the life cycle of the watcher. Not to the contents of the proxy/dict.

Adding a main dep seems to me like a kind of shotgun approach and negates the whole reason why you would have keydeps, right?

Nice little brain twister this! 😄

Korijn:
I'll go with it, don't have the headspace to give this my full attention :)

observ/traps.py
Korijn: This is not true though, it's here:

https://github.com/fork-tongue/observ/blob/eb08db903210f85ffc9461e3d3f16e720e1bae4f/observ/proxy_db.py#L82

And it is notified here:

https://github.com/fork-tongue/observ/blob/eb08db903210f85ffc9461e3d3f16e720e1bae4f/observ/traps.py#L122C13-L122C34

To me it feels like the real bug is that `is_new` apparently does not evaluate to `True` in the circumstances your failing test reproduces:

https://github.com/fork-tongue/observ/blob/eb08db903210f85ffc9461e3d3f16e720e1bae4f/observ/traps.py#L110

berendkleinhaneveld: Sorry, I'm talking about 'depend': it will not call `keydeps[key].depend()` (which happens in the `read_key_trap`). I'm wondering why we even try to delete all keydep + subs in the delete key trap anyways?

Korijn: Yeah so here's what I'm confused about (I could look into it a bit more in the morning). After the delete key trap has run, the state should be the same as at the point where the test starts. So when you assign True to the key again, when the write_key_trap runs, it should create a new Dep and next it should subscribe immediately as well, just like it did initially. But it doesn't. This seems wrong.

berendkleinhaneveld: That is because the watcher is only subscribed when the read trap is triggered. But that doesn't happen anymore.

berendkleinhaneveld: Fixed by just not cleaning up anything (like Vue): #159

PR #159, opened 2026-02-03
Don't clean up keydeps at all, like Vue

### LINKED ISSUE #134: Calling pop with a default arg results in KeyError

berendkleinhaneveld:
Just to share a bit of context: I've looked through the [reactivity core](https://github.com/vuejs/core/tree/main/packages/reactivity) package and tried to map the types in observ to the types in Vue. The `targetMap` in Vue maps to the `proxy_db` (mapping objects to Deps) and the I've tried to figure out what Vue does when a key is removed, but the only calls `trigger` (`notify` in observ). No explicit cleanup is done: it simply depends on the objects' life cycle.

This PR is the answer to the conversation in #153 .

### DIFF
--- a/observ/traps.py
+++ b/observ/traps.py
@@ -135,7 +135,6 @@ def delete_trap(method, obj_cls):
         attrs["dep"].notify()
         for key in self._orphaned_keydeps():
             attrs["keydep"][key].notify()
-            del attrs["keydep"][key]
         return retval
 
     return trap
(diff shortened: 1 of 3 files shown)


Korijn (approved):

Korijn:
I wonder if we even need to trap delete methods anymore?

PR #162, opened 2026-02-22
Add example using rendercanvas

almarklein:


almarklein:
Made this part of #163

Issue #161, opened 2026-02-09
Update PySide integration with asyncio


berendkleinhaneveld:
The current way of asyncio integration triggers a warning about using deprecated asyncio methods. 
Note that changing this might require dropping support for older python versions though (when used in combination with PySide).

During investigation I also noticed that the qt example is not working properly anymore for PySide 6.8 and up. Would be nice to fix that as well.

berendkleinhaneveld:
Fixed by #164

PR #163, opened 2026-02-22
Add scheduler.register_rendercanvas()

almarklein:
This makes it a bit more obvious how to hook up with rendercanvas. Makes anything that uses Pygfx+observ just a bit easier.

(If both this and #162 is accepted, we should update the example to use this new method.)

### DIFF
--- a/observ/init.py
+++ b/observ/init.py
@@ -3,12 +3,15 @@ import asyncio
 from .scheduler import scheduler
 
 
-def init(mode="asyncio"):
+def init(mode="asyncio", loop=None):
     if mode == "qt":
-        scheduler.register_qt()
+        scheduler.register_qt(loop)
 
-    if mode == "asyncio":
-        scheduler.register_asyncio()
+    elif mode == "asyncio":
+        scheduler.register_asyncio(loop)
+
+    elif mode == "rendercanvas":
+        scheduler.register_rendercanvas(loop)
 
 
 def loop_factory():
--- a/observ/init.py
+++ b/observ/init.py
@@ -5,7 +5,7 @@ from .scheduler import scheduler
 
 def init(mode="asyncio", loop=None):
     if mode == "qt":
-        scheduler.register_qt(loop)
+        scheduler.register_qt()
 
     elif mode == "asyncio":
         scheduler.register_asyncio(loop)
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -96,6 +96,14 @@ class Scheduler:
         self.timer.setInterval(0)
         self.register_request_flush(self.timer.start)
 
+    def register_rendercanvas(self, loop):
+        """
+        Utility function for integration with rendercanvas loop objects
+        """
+        if not hasattr(loop, "call_soon"):
+            raise TypeError("Given loop object does not have a call_soon method.")
+        self.register_request_flush(lambda: loop.call_soon(self.flush))
+
     def flush(self):
         """
         Flush the queue to evaluate all queued watchers.
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -106,9 +106,10 @@ class Scheduler:
         """
         Utility function for integration with rendercanvas loop objects
         """
-        if not hasattr(loop, "call_soon"):
+        needed_methods = {"call_soon", "call_soon_threadsafe"}
+        if not all(hasattr(loop, m) for m in needed_methods):
             raise TypeError(
-                f"Given loop object does not have a call_soon method: {loop!r}"
+                f"Given loop object does not have all needed methods: {needed_methods!r}"
             )
         # Since rc loop objects look similar to asyncio, we can reuse the method
         self.register_asyncio(loop)
--- a/observ/scheduler.py
+++ b/observ/scheduler.py
@@ -54,14 +54,17 @@ class Scheduler:
 
     def register_asyncio(self, loop=None):
         """
-        Utility function for integration with asyncio
+        Utility function for integration with asyncio.
+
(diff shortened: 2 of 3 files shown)


Korijn:
Can you also see if you can add this to `observ/init.py`?

berendkleinhaneveld:
Would it make sense to call it: `register_loop` or something more descriptive? Now that I'm thinking about it: we could also give the asyncio an optional loop argument instead of adding this method?

almarklein:
> Can you also see if you can add this to `observ/init.py`?

Yeah, but this new method does require a `loop` argument; I added it as an optional arg to `init()`, and also to the `register_` methods for consistency.

> Now that I'm thinking about it: we could also give the asyncio an optional loop argument instead of adding this method?

It would work, because rendercanvas loop and asyncio loop objects both have a `call_soon()` and `call_later()`, but it may not be obvious to the user, also because rendercanvas could also run on trio/qt or something else.  I figured that from the p.o.v. of the user it makes sense to make it explicit.

----

Maybe a bit of background. I had code like this:
```py
from rendercanvas.auto import RenderCanvas, loop
from observ import scheduler, reactive

scheduler.register_asyncio()

state = reactive(...)
_watcher = watch(my_reactive_func)

loop.run()

```

This worked for a while but stopped working recently, and I don't even know why it worked earlier 😆. What happens is that when the watcher is created, observ uses `asyncio.get_event_loop()`. But the event loop that this function returns is never run; `loop.run()`  uses `asyncio.run()` (for the asyncio loop backend), which always creates a new loop.

----

The `asyncio.get_event_loop()` is weird ...

```py
import asyncio

print("1", id(asyncio.get_event_loop()))
print("2", id(asyncio.get_event_loop()))

async def main():
    print("main", id(asyncio.get_running_loop()))
    
asyncio.run(main())

print("3", id(asyncio.get_event_loop()))
```

Produces:
```
<tmp 3>:3: DeprecationWarning: There is no current event loop
  print("1", id(asyncio.get_event_loop()))
1 4394534288
2 4394534288
main 4393890384
Traceback (most recent call last):
  File "<tmp 3>", line 13, in <module>
    print("3", id(asyncio.get_event_loop()))
                  ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/almar/.local/share/uv/python/cpython-3.13.5-macos-aarch64-none/lib/python3.13/asyncio/events.py", line 716, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
                       % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

Why is there a deprecation warning. Why does `asyncio.run()` use that "default" loop? Why does `get_event_loop()` raise work on init, but raises an error after `run()` is called ...

Korijn (approved):
LGTM!

observ/scheduler.py
```diff
@@ -52,20 +52,25 @@ def request_flush_asyncio(self):
...
         Utility function for integration with asyncio
         """
-        self.register_request_flush(self.request_flush_asyncio)
+        if loop is not None:
+            self.register_request_flush(lambda: loop.call_soon(self.flush))
+        else:
+            self.register_request_flush(self.request_flush_asyncio)
 
-    def register_qt(self):
+    def register_qt(self, loop=None):
```
berendkleinhaneveld: I would just leave out the loop here, so that it is clear that this method doesn't take any loop.

observ/scheduler.py
```diff
@@ -96,6 +101,14 @@ def register_qt(self):
         self.timer.setInterval(0)
         self.register_request_flush(self.timer.start)
 
+    def register_rendercanvas(self, loop):
+        """
+        Utility function for integration with rendercanvas loop objects
+        """
+        if not hasattr(loop, "call_soon"):
+            raise TypeError(f"Given loop object does not have a call_soon method: {loop!r}")
+        self.register_request_flush(lambda: loop.call_soon(self.flush))
```
berendkleinhaneveld: Here, I would call `self.register_asyncio(loop)`.

berendkleinhaneveld (commented):
To me it also makes sense to combine this PR with #162, so that it's possible to try out directly. I did add the example locally, to try it out and updated it to call `init("rendercanvas", loop)`: worked nicely!

observ/scheduler.py
almarklein: done

observ/scheduler.py
almarklein: yeah, let's de-duplicate that. Added a comment to explain.

observ/scheduler.py
```diff
@@ -96,6 +102,17 @@ def register_qt(self):
         self.timer.setInterval(0)
         self.register_request_flush(self.timer.start)
 
+    def register_rendercanvas(self, loop):
+        """
+        Utility function for integration with rendercanvas loop objects
+        """
+        if not hasattr(loop, "call_soon"):
```
berendkleinhaneveld: In #161 I made the flush call the 'call_soon_threadsafe' method, to make sure the effects are run in the main thread (useful for Qt/UI integration). So better to check for that here now

observ/init.py
```diff
@@ -3,12 +3,15 @@
 from .scheduler import scheduler
 
 
-def init(mode="asyncio"):
+def init(mode="asyncio", loop=None):
     if mode == "qt":
-        scheduler.register_qt()
+        scheduler.register_qt(loop)
```
berendkleinhaneveld: Ah, now I think I got it why you made it similar to the other methods... 
Anyways, the loop argument here should be removed.

berendkleinhaneveld (approved):

berendkleinhaneveld:
As soon as you fix the linting, we can merge this!

PR #164, opened 2026-03-30
Use 'init' method and fix progress bar display

berendkleinhaneveld:
Apparently, calling anonymous functions / closures as slots from within a QThread does not work well. When the callbacks were changed to be class methods, then the progress bar got displayed nicely again. I did notice that the callback got called properly, but it seems that it just doesn't trigger any change.

I'm opening this as draft, want to see if I can better understand this before closing it.

### DIFF
--- a/examples/observe_qt.py
+++ b/examples/observe_qt.py
@@ -107,16 +107,16 @@ class Controls(QWidget):
 
         self.thread.start()
 
+        def progress(x):
+            self.state["progress"] = x
+
+        def bump(x):
+            self.state["clicked"] += x * 0.5
+
         self.button.setEnabled(False)
         self.thread.finished.connect(lambda: self.button.setEnabled(True))
-        self.worker.result.connect(self.on_result)
-        self.worker.progress.connect(self.on_progress)
-
-    def on_progress(self, x):
-        self.state["progress"] = x
-
-    def on_result(self, x):
-        self.state["clicked"] += x * 0.5
+        self.worker.result.connect(bump)
+        self.worker.progress.connect(progress)
 
     def on_reset_clicked(self):
         self.state["clicked"] = 0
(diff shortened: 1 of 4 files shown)


Korijn:
Modifying Qt UI state from non-main threads doesn't work. Normally speaking you design two step callback chains in Qt: main thread triggers a callback -> worker thread does some work and triggers a callback -> main thread updates the UI.

Maybe by using class methods you are implicitly scheduling work on the main thread?

berendkleinhaneveld:
Yeah, apparently when PySide6 detects a QObject, then it will post the event on its associated event queue. A lambda does not have an event queue, so it basically sent to the void.

Claude spewed out this chain of events:

> Here's the chain when state["progress"] = x runs on the worker thread:
>  1. `dep.notify()` → watchers queued in the scheduler
>  2. `request_flush_asyncio()` → `asyncio.get_event_loop().call_soon(self.flush)`
>  3. `call_soon` creates a QAsyncioHandle with `is_threadsafe=False`
>  4. That calls `QTimer.singleShot(0, flush)` — which creates the timer on the calling (worker) thread
>  5. The worker thread's event loop isn't running (it's blocked in sleep), so the timer never fires

That last part might be BS, but it is went on by mentioning something interesting about the `request_flush_asyncio`:

> If `call_soon_threadsafe` were used instead, the handle would be created with `is_threadsafe=True`, which uses the `QTimer.singleShot(timeout, self._loop, func)` overload that posts to the main thread's event loop.

Basically, if we would use `call_soon_threadsafe`, we could ensure that flushes are always done in the main thread. I think that might be nice. What do you think?

Korijn:
Yeah that sounds right to me. Using singleshot is a pattern commonly used in Qt apps to schedule to the main thread. You can also use signals and slots (yuck). I vaguely remember from the very early CG days there is a way to specifically indicate some callback should be scheduled on the main thread but I can't for the life of me find the related documentation.

Korijn (approved):

berendkleinhaneveld:
I might update the testsuite as well to not use the deprecated way of setting a loop policy, but depending on how much effort it is I might also do that in another PR.

Done!

PR #165, opened 2026-05-29
Fix deepcopy of Proxy objects creating unregistered zombie proxies

berendkleinhaneveld:
When `copy.deepcopy()` is called on an `observ` Proxy, Python's default deepcopy copies the slots without calling `__init__`. This means `proxy_db.reference()` is never called, producing a "zombie proxy" whose target is unknown to `proxy_db`. Any subsequent access raises `KeyError` in `proxy_db.attrs()`.

This surfaces when `reliev/patchdiff` call `deepcopy` on patch values that are Proxy objects during undo/redo.

This would also make some code in `patchdiff` a bit simpler, by just being able to call `deepcopy` on the state that it operates on.

### DIFF
--- a/observ/proxy.py
+++ b/observ/proxy.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from copy import copy, deepcopy
 from functools import partial
 from typing import Generic, Literal, TypedDict, TypeVar, cast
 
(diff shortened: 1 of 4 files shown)


berendkleinhaneveld:
I think this changes the behavior enough to warrant a minor version bump.

Korijn:
Is this because of the deepcopy calls that were added in one of the PRs this week?

berendkleinhaneveld:
Yes, the deepcopy calls in patchdiff. I think those are still required. I could of course try and update the code in patchdiff to check whether it is a proxy and then call `to_raw` and then `deepcopy` it to make sure no modifications will end up in the patches, but this seems like the better solution. And it also allows to simplify some code in patchdiff, as shown in the linked PR I just opened in patchdiff..

Korijn (approved):

PR #170, opened 2026-07-02
Traverse raw targets instead of proxy iterators

berendkleinhaneveld:
Part of the performance series following #166. Independent of #167/#168/#169 (different code regions; #167/#168 touch `proxy.py` which this PR only imports).

Deep watchers re-run `traverse` on every evaluation, and traversing through the proxies' trapped iterators created a generator plus a child proxy for **every visited value**, every time. This was the dominant cost of deep watching.

## Approach

`traverse` now walks the raw targets and registers a dependency on each container's `dep` directly. This is equivalent to the old behavior: every mutation trap notifies the dep of the container it mutates (key writes notify both the keydep and the container dep). Raw containers nested inside proxied targets are registered in the proxy_db along the way — the old implementation did the same implicitly by proxying them during iteration — so mutations through (future) proxy paths keep notifying the watcher.

A `tracked` flag on the traversal stack preserves the existing semantics exactly:
- children of **shallow** proxies are not tracked (matching shallow iterators, which yield raw values),
- raw containers not reachable through any proxy are not tracked (there is no proxy through which they can be mutated),
- proxies encountered anywhere always track their own dep.

Verified behavior parity for: nested never-read containers, containers added after the watch, shallow proxies, cyclic structures, and proxies inside raw containers.

## Bug surfaced along the way

The dict/list/set proxy types were registered in `TYPE_LOOKUP` as a side effect of `watcher.py`'s imports — with those imports removed, `reactive({})` returned a plain dict in a fresh interpreter. The top-level package now imports the proxy modules explicitly, with a fresh-interpreter regression test (inside the test suite this is masked, because other test modules import the proxy modules directly).

## Benchmark results (mean, local run, vs master)

| Benchmark | master | this PR | speedup |
|---|---|---|---|
| `traverse_deep` | 363 µs | 46 µs | 7.9x |
| `traverse_wide` | 4.28 ms | 0.51 ms | 8.3x |
| `traverse_mixed` | 211 µs | 27 µs | 7.7x |
| `traverse_matrix` | 1.98 ms | 0.27 ms | 7.3x |
| `traverse_stress_large_shallow` | 727 µs | 105 µs | 6.9x |
| `deep_watch_mutate_leaf[10]` | 347 µs | 74 µs | 4.7x |
| `deep_watch_mutate_leaf[1k]` | 35.8 ms | 6.8 ms | 5.3x |

(The standalone `traverse_*` benchmarks run outside a watcher and so skip dep registration entirely; the `deep_watch_mutate_leaf` numbers show the full in-watcher path.)

Full test suite passes (164 passed, 1 xfailed).

### DIFF
--- a/observ/__init__.py
+++ b/observ/__init__.py
@@ -3,6 +3,8 @@ from importlib.metadata import version
 __version__ = version("observ")
 
 
+# Importing the proxy modules registers their types in TYPE_LOOKUP
+from . import dict_proxy, list_proxy, set_proxy
 from .init import init, loop_factory
 from .proxy import (
     reactive,
--- a/observ/watcher.py
+++ b/observ/watcher.py
@@ -16,10 +16,9 @@ from typing import Any, Callable, Generic, Optional, TypeVar, Union
 from weakref import WeakSet, ref
 
 from .dep import Dep
-from .dict_proxy import DictProxyBase
-from .list_proxy import ListProxyBase
+from .proxy import Proxy, proxy
+from .proxy_db import proxy_db
 from .scheduler import scheduler
-from .set_proxy import SetProxyBase
 
 T = TypeVar("T")
 Watchable = Union[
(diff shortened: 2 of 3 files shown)


observ/__init__.py
```diff
@@ -3,6 +3,8 @@
 __version__ = version("observ")
 
 
+# Importing the proxy modules registers their types in TYPE_LOOKUP
+from . import dict_proxy, list_proxy, set_proxy
```
berendkleinhaneveld, line 7: We'll probably want to revert this

berendkleinhaneveld: Actually, this works OK. Seems to do the job well.

berendkleinhaneveld:
Before baking a release with these changes I'll perform an end-to-end test with these changes as well, to make sure there are no regressions.

berendkleinhaneveld:
It all seems to work properly!

PR #171, opened 2026-07-02
Create keydeps lazily plus assorted trap/dep micro-optimizations

berendkleinhaneveld:
Last of the planned performance series following #166. Branched off master, no textual overlap with #167/#168/#169/#170 (touches `write_key_trap`/`delete_key_trap`/`iterate_trap`, while #169 rewrites `write_trap` — different functions in the same file, so git merges them cleanly in any order).

## Lazy keydep creation (the main change)

Proxying a dict eagerly created a `Dep` for every key, making proxy creation O(n) — ~20 ms for a 100k-key dict. Keydeps are now created lazily when a key is read with dependency tracking active (this matches Vue, which creates deps on track). Writes no longer create subscriber-less keydeps either: a keydep without subscribers serves no purpose, and subscribing requires reading the key, which creates it. Once created, keydeps still live forever (per "Don't clean up keydeps at all, like Vue").

Consequences in the traps:
- `write_key_trap` derives `is_new` from the target instead of from keydep presence (which was only equivalent under eager creation) and notifies the keydep only if it exists.
- `delete_key_trap` tolerates a missing keydep.

The white-box tests that subscribed mocks to eagerly created keydeps now seed those keydeps explicitly; `test_deps_delete` additionally documents the lazy-creation lifecycle. Its core assertion — keydeps are never removed on delete — is unchanged.

## Smaller items

- Hoist the `method == "items"` / `method == "setdefault"` string comparisons out of the trap closures (known at trap construction time).
- `operator.attrgetter` instead of a lambda for the sort keys in `Dep.notify` and `Scheduler.flush`.
- Look up the proxy_db entry once in `reference()`/`dereference()` instead of once per access.

## Benchmark results (mean, local run, vs master)

| Benchmark | master | this PR |
|---|---|---|
| `proxy_creation_dict[100k]` | 20.1 ms | 26 µs |
| `proxy_creation_dict[1k]` | 84 µs | 16 µs |
| `proxy_creation_dict[10]` | 27 µs | 15 µs |
| `write_dict_setitem`, `write_dict_new_key` | ~0.9–1.0 µs | ~7% faster |
| `notify_watchers` | — | no measurable change (the sort is a small share of the notify pipeline) |

Full test suite passes (163 passed, 1 xfailed).

### DIFF
--- a/observ/dep.py
+++ b/observ/dep.py
@@ -5,8 +5,11 @@ are attached to observable datastructures.
 
 from __future__ import annotations
 
+from operator import attrgetter
 from typing import ClassVar
 from weakref import WeakSet
+
+sub_id = attrgetter("id")
 
 
 class Dep:
@@ -36,5 +39,5 @@ class Dep:
         # because a weakset must acquire a lock on its
         # weak references before iterating
         if self._subs:
-            for sub in sorted(self._subs, key=lambda s: s.id):
+            for sub in sorted(self._subs, key=sub_id):
                 sub.update()
--- a/observ/proxy_db.py
+++ b/observ/proxy_db.py
@@ -75,19 +75,23 @@ class ProxyDb:
         target = proxy.__target__
         obj_id = id(target)
 
-        if obj_id not in self.db:
+        entry = self.db.get(obj_id)
+        if entry is None:
             attrs = {}
             if isinstance(target, dict):
                 attrs["dep"] = Dep()
-                attrs["keydep"] = {key: Dep() for key in target.keys()}
+                # keydeps are created lazily: when a key is read
+                # (with dependency tracking active) or written
+                attrs["keydep"] = {}
             elif isinstance(target, (list, set)):
                 attrs["dep"] = Dep()
-            self.db[obj_id] = {
+            entry = {
                 "target": target,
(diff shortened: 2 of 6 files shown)


berendkleinhaneveld:
Nice, for creation of really big dicts:
```
----------------------------- benchmark 'proxy_creation_dict': 6 tests -----------------------------
Name (time in us)                                       Mean                StdDev            Rounds
----------------------------------------------------------------------------------------------------
test_proxy_creation_dict[10] (0001_master)           26.2136 (1.0)          3.5271 (1.58)         30
test_proxy_creation_dict[10] (NOW)                   30.9743 (1.18)         2.2284 (1.0)          30
test_proxy_creation_dict[1k] (NOW)                   31.4177 (1.20)         4.8899 (2.19)         30
test_proxy_creation_dict[100k] (NOW)                 46.6942 (1.78)         5.9395 (2.67)         30
test_proxy_creation_dict[1k] (0001_master)          161.8048 (6.17)        14.1141 (6.33)         30
test_proxy_creation_dict[100k] (0001_master)     23,752.7727 (906.13)   1,541.9236 (691.95)       30
----------------------------------------------------------------------------------------------------
```

PR #177, opened 2026-07-08
Manage proxy_db lifetimes with reference counting instead of a gc hook

Korijn:
## Motivation

observ's whole cleanup story hinged on a garbage-collector hook: `proxy_db` held a **strong** reference to every wrapped target, keyed on `id(target)`, and a `gc.callbacks` hook combined with `sys.getrefcount()` inspection (in the hook and in `Proxy.__del__`) decided when entries could be dropped. Consequences:

- Entries for targets still referenced elsewhere were only released on a *full gen-2 collection*, which is rare — so memory could grow substantially in between.
- With `gc.disable()` (not uncommon in latency-sensitive apps), those entries leaked forever.
- Every full collection paid for a Python-land scan of the whole db.
- Correctness rested on fragile refcount arithmetic ("if 2: we are the last to hold a reference!").

## Design

Invert the ownership so that plain reference counting does all cleanup, and the registry never needs to be scanned:

- The per-target entry is now a **`TargetDep`** — the `Dep` for the container itself, which also owns the target, the per-key deps, and (weakly) the proxies that wrap it. Owning the target is what keeps `id(target)` keying safe: an id can only be recycled after its `TargetDep` is gone and has removed itself from the registry.
- Every `Proxy` holds a strong reference to its `TargetDep` (`__dep__`).
- Per-key deps are **`KeyDep`** instances that hold their owning `TargetDep`. Watchers already hold strong references to the deps they depend on, so an entry stays alive for exactly as long as a proxy exists *or* a subscribed watcher needs its dep identity — the transient-proxy notification path keeps working (covered by a new test).
- `proxy_db.db` maps `id(target)` → *weakref* to the `TargetDep`; entries remove themselves via weakref callbacks (with an identity guard against stale callbacks).

No gc hook, no refcount inspection, no `Proxy.__del__`, and deliberately **no reference cycles** among `TargetDep`/`KeyDep`/proxies — cleanup is deterministic and immediate, and works with the garbage collector fully disabled. Keydeps become weakly registered: a keydep without subscribers has nobody to notify, so it can be recreated freely on the next tracked read (this also removes the vestigial keydep creation in the dict write trap).

## Performance

Traps now reach their deps through a slot (`self.__dep__`) instead of a global-db `id()` dict lookup, entry creation no longer constructs `WeakValueDictionary`s (keydeps are materialized lazily, proxies are a plain dict of weakrefs), and the gen-2 gc callback overhead is gone entirely. Spot check (min of 5 timeit repeats, Python 3.11):

| op | master | this PR |
|---|---|---|
| keyed/nested reads | 6.5 µs | 5.0 µs |
| keyed write (watched) | 4.6 µs | 4.2 µs |
| `reactive()` creation | 4.1 µs | 1.9 µs |
| deep-watched list write | 144 µs | 134 µs |

## Behavioral changes

- A registry entry is now released as soon as the last proxy dies and no watcher depends on the target — previously it was kept while the *raw target* was alive. Since dep identity only matters to subscribers, this is not observable, but white-box lifecycle tests were updated accordingly.
- `proxy_db.attrs(proxy)` is replaced by direct `proxy.__dep__` / `proxy.__dep__.keydeps` access.
- The `clear_proxy_db` conftest fixture is no longer needed and was removed.
- Verified: 10k create/mutate/destroy churn iterations with `gc.disable()` leak zero registry entries, and `gc.callbacks` stays empty.

Full test suite passes (including qt group, run offscreen), plus ruff check/format.

https://claude.ai/code/session_01T4cFbKSCxjGabtJbB9e5H2

---
_Generated by [Claude Code](https://claude.ai/code/session_01T4cFbKSCxjGabtJbB9e5H2)_

### DIFF
--- a/observ/dict_proxy.py
+++ b/observ/dict_proxy.py
@@ -1,5 +1,4 @@
 from .proxy import TYPE_LOOKUP, Proxy
-from .proxy_db import proxy_db
 from .traps import construct_methods_traps_dict, trap_map, trap_map_readonly
 
 dict_traps = {
@@ -54,7 +53,10 @@ class DictProxyBase(Proxy[dict]):
     __slots__ = ()
 
     def _orphaned_keydeps(self):
-        return set(proxy_db.attrs(self)["keydep"].keys()) - set(self.__target__.keys())
+        keydeps = self.__dep__.keydeps
+        if keydeps is None:
+            return set()
+        return set(keydeps.keys()) - set(self.__target__.keys())
 
 
 def readonly_dict_proxy_init(self, target, shallow=False, **kwargs):
--- a/bench/test_creation.py
+++ b/bench/test_creation.py
@@ -1,9 +1,8 @@
 """
 Benchmarks for the cost of creating proxies and watchers.
 
-Creating a proxy for a dict registers the target in the proxy_db and
-eagerly creates a Dep for every key, so proxy creation cost scales
-with the number of keys.
+Creating a proxy registers the target in the proxy_db; deps for the
+keys of a dict are created lazily when they are read or written.
 """
 
 import gc
@@ -26,9 +25,8 @@ def noop():
 def test_proxy_creation_dict(benchmark, size):
     def setup():
         # Collect garbage in between rounds (outside of the measured
-        # code) so that proxy_db entries for the previous rounds are
-        # cleaned up deterministically instead of adding gc noise to
-        # the measurement
+        # code) so that garbage from the previous rounds doesn't add
+        # gc noise to the measurement
         gc.collect()
         return ({f"key_{i}": i for i in range(size)},), {}
 
--- a/docs/guide/gotchas.md
+++ b/docs/guide/gotchas.md
@@ -2,9 +2,9 @@
 
 ## Keep no references to raw data
 
-Observ keeps references to the object passed to `reactive()` in order to keep track of dependencies and proxies for that object. When the object that is passed into `reactive()` is not managed by other code, observ cleans up its references automatically when the proxy is destroyed. However, if there is another reference to the original object, observ will only release its own reference when the garbage collector runs and all other references to the object are gone.
+Observ keeps a reference to the object passed to `reactive()` in order to keep track of dependencies and proxies for that object. That reference is released automatically — through regular reference counting, without relying on the garbage collector — as soon as the last proxy for the object is destroyed and no watcher depends on it anymore.
 
-For this reason, the **best practice** is to keep **no references** to the raw data, and instead work with the reactive proxies **only**:
+Even so, the **best practice** is to keep **no references** to the raw data, and instead work with the reactive proxies **only**: mutations on the raw data itself bypass the proxies and are therefore not observable.
 
 ```python
 # Good: no reference to the raw dict survives
--- a/observ/proxy.py
+++ b/observ/proxy.py
@@ -13,8 +13,12 @@ class Proxy(Generic[T]):
     """
     Proxy for an object/target.
 
-    Instantiating a Proxy will add a reference to the global proxy_db and
-    destroying a Proxy will remove that reference.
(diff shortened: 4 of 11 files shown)


observ/dict_proxy.py
```diff
@@ -54,7 +53,10 @@ class DictProxyBase(Proxy[dict]):
     __slots__ = ()
 
     def _orphaned_keydeps(self):
-        return set(proxy_db.attrs(self)["keydep"].keys()) - set(self.__target__.keys())
+        keydeps = self.__dep__.keydeps
+        if keydeps is None:
+            return set()
+        return set(keydeps.keys()) - set(self.__target__.keys())
```
berendkleinhaneveld, line 59:
```
>>> a = {"a": "a"}
>>> b = {"b": "b"}
>>> a.keys()
dict_keys(['a'])
>>> b.keys()
dict_keys(['b'])
>>> a.keys() - b.keys()
{'a'}
>>> type(a.keys() - b.keys())
<class 'set'>
```
Just verified that this works already in Python 3.10, so this code could just be:


```suggestion
        return keydeps.keys() - self.__target__.keys()
```

berendkleinhaneveld: Ah, wait: I need to check with the proper types :')

berendkleinhaneveld: Ok, this does not work for WeakValueDictionaries

berendkleinhaneveld (approved):
Love it!

Korijn:
## Heads-up: benchmark gate regression on `test_watcher_creation`

The Benchmarks job fails with two regressions vs `master`:

- **`test_watcher_creation`: +57%** — the real problem
- `test_read_list_iterate[reactive]`: +5.5% — right at the threshold, looks like variance (lists have no keydeps)

### Root cause

`test_watcher_creation` repeatedly creates and discards a watcher over the same `reactive({"count": 0})`. On `master`, the keydep for `"count"` was a plain `Dep` held **strongly** in the entry's keydep dict, so it was created once and reused every round.

In this PR keydeps live in a `WeakValueDictionary` and are only kept alive by a *subscribed watcher*. Between rounds the previous watcher is dropped, so its keydep dies and is recreated — a fresh `KeyDep` + weakref-with-callback **every single round**. Instrumented: 2000 create/destroy cycles → 2000 `KeyDep` allocations (should be ~1). Pinning the keydep so it's reused makes creation **1.77× faster**, which accounts for the whole regression.

This weakness is deliberate — it's what lets an entry be reclaimed by pure reference counting (no cycles, works even with `gc.disable()`). Fixing the churn means keydeps have to **persist and be reused**, and that reintroduces a design fork:

### Options

1. **Strong keydeps + keep `KeyDep.owner` (simple).** ~3 lines. Restores perf exactly. Introduces an ordinary `TargetDep ↔ KeyDep` cycle reclaimed by CPython's *standard* cyclic collector. Still removes the `gc.callbacks` **hook** entirely (the actual goal), but drops the "no cycles / gc-disable-safe" bonus — I'd update the PR text to match.

2. **Strong plain keydeps + watcher-side owner retention.** Keeps full reference-counting cleanup (no cycles, gc-disable-safe) and every current claim. More invasive: the `Watcher` separately retains each owning `TargetDep`, touching its dep bookkeeping, cleanup, and `depend()` propagation.

3. **Leave keydeps weak, claw the 57% back elsewhere in the creation path.** Investigating now — if `Watcher.__init__` overhead can be trimmed enough to bring the net under the 5% gate, we keep the current lifetime model unchanged.

Leaning toward exploring (3) first since it preserves the cleanest semantics; will report back. Happy to take (1) if you'd rather keep it minimal.

---

berendkleinhaneveld:
I propose to merge and let this `test_watcher_creation` regression be. Makes sense there is a bit of overhead in this specific case but I much prefer this architecture.

Korijn:
I agree! 💯

PR #183, opened 2026-07-09
Fully implement modern type hinting, checked with ty in CI (closes #114)

### LINKED ISSUE #114: Type hinting improvements
There are some items left from #111:

* https://github.com/fork-tongue/observ/pull/111#discussion_r1392316931
* We could also add mypy to the dev toolchain perhaps.
* Watchable callables actually don't have to return anything.
* We could use `|` instead of `Union`

Korijn:
Modernize and complete the type hints across the whole package, and
guard against drift by type checking with ty (Astral's type checker)
in CI and pre-commit.

Closing the remaining items from #111/#114:

* PEP 604 unions ('X | Y') everywhere instead of Union/Optional. The
  Watchable and WatchCallback aliases live in a TYPE_CHECKING block so
  the '|' syntax never runs on Python 3.9.
* The casts in to_raw are needed (the type system cannot know that
  rebuilding a container from unproxied items preserves its type);
  this is now documented in place. In the hot proxy() path, casts are
  replaced with zero-cost Any-typed locals since typing.cast is a real
  function call at runtime.
* Watchable callables don't have to return anything: T binds to None
  for effect-style functions, documented on the alias.

Other improvements that fall out of accurate typing:

* Ship a py.typed marker (PEP 561) and the Typing :: Typed classifier
  so type checkers pick up the annotations downstream.
* watch_effect is a real function instead of a functools.partial, so
  it gets an accurate generic signature, a docstring, and mkdocstrings
  rendering in the API reference.
* readonly/shallow_reactive/shallow_readonly are real functions
  instead of partials, for the same reason.
* computed is typed with overloads returning a Computed protocol, so
  the __watcher__ attribute is part of the public type.
* ref() no longer needs the try/except TypedDict definition dance:
  the generic Ref TypedDict is typing-only, all Python versions share
  one runtime implementation.
* is_bound_method is a TypeIs guard, narrowing to MethodType.
* Fixed annotation bugs the type checker found: _run_callback returns
  a value but was annotated -> None, Dep._subs was typed WeakSet but
  assigned None, and Watcher.on_created/on_destroyed are ClassVars.
* Scheduler grows a SupportsCallSoon protocol for the loop objects
  accepted by register_asyncio/register_rendercanvas.
* Internal helper signatures (traps, proxy_db, scheduler) are fully
  annotated; the string-and-noqa forward references are gone.

The traceback inspection in _run_callback now explicitly deletes its
local reference to the traceback: storing it in a local created a
frame <-> traceback reference cycle that kept watchers (and everything
they reference) alive until the next gc collection, breaking the
refcount-based lifetime that the weakref machinery (and the tests)
rely on.

Benchmarks were run interleaved against master with the repo's
compare_runs.py guard: no benchmark consistently regressed.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016fDq6fyKg6QscqyyN6CcwC

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,21 @@ jobs:
       - name: Format
         run: uv run --no-sync ruff format --check .
 
+  typecheck:
+    name: Typecheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install uv
+        uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
+      - name: Install only ty
+        run: uv sync --only-group ty
+      - name: Typecheck
+        run: uv run --no-sync ty check
+
   test:
     name: Test on ${{ matrix.name }}
     runs-on: ubuntu-latest
@@ -93,7 +108,7 @@ jobs:
   publish:
     name: Publish to Github and Pypi
     runs-on: ubuntu-latest
-    needs: [lint, test, build]
+    needs: [lint, typecheck, test, build]
     if: success() && startsWith(github.ref, 'refs/tags/v')
     environment:
       name: pypi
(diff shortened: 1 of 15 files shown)


berendkleinhaneveld (approved):
Awesome, looks good! 
The only thing that surprised me was using class level typing for attributes, but that should be fine.

PR #184, opened 2026-07-09
Runtime cleanups split out from the type hinting work

Korijn:
Part 1 of 2 for #114: the behavioral/code changes, split out from the pure typing changes (which stack on top of this PR) so each can be reviewed in isolation.

## Changes

* **`watch_effect` is a real function** instead of a `functools.partial`, so it has an inspectable signature, a docstring, and mkdocstrings rendering in the API reference (the hand-written docs section is replaced by the generated one).
* **`readonly`/`shallow_reactive`/`shallow_readonly` are real functions** instead of partials, for the same reason (partials also type-check poorly, which matters for part 2).
* **`ref()` has a single implementation**: the generic `Ref` TypedDict is typing-only (it requires Python 3.11 at runtime, hence the old try/except definition dance), and at runtime a `Ref` was always just a plain (proxied) dict anyway.
* **`init(mode="rendercanvas")` without a loop** raises a clear `TypeError` instead of failing inside `register_rendercanvas`.
* **`loop_factory()`** looks up `asyncio.eager_task_factory` once with `getattr` instead of `hasattr` + attribute access.

No public API changes beyond the partial→function swaps (call sites are unaffected; keyword overrides that the partials accidentally allowed, like `watch_effect(fn, immediate=True)`, are no longer possible).

**Note on `_run_callback`:** an earlier version of this PR also "fixed" a reference-cycle leak in the traceback-introspection code there. On review, that leak didn't exist in the original code — it relied on Python's implicit `del e` at the end of `except TypeError as e:`, which already prevents the frame↔traceback cycle. My rewrite introduced the leak itself by binding `e.__traceback__` to its own local (not covered by that implicit cleanup). `_run_callback` is therefore left untouched here; part 2 handles the one line that needs a type-checker accommodation, via a narrow ignore comment rather than a control-flow change.

https://claude.ai/code/session_016fDq6fyKg6QscqyyN6CcwC

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -52,13 +52,7 @@ Combination of `shallow_reactive` and `readonly`.
 
 ::: observ.watcher.watch
 
-### `watch_effect`
-
-```python
-watch_effect(fn: Callable[[], Any]) -> Watcher
-```
-
-Runs `fn` immediately to collect its dependencies and re-runs it (via the scheduler) whenever they change. Equivalent to `watch(fn, deep=True)` without a callback. See [Watchers](../guide/watchers.md#watch_effect).
+::: observ.watcher.watch_effect
 
 ::: observ.watcher.computed
 
--- a/observ/init.py
+++ b/observ/init.py
@@ -18,6 +18,8 @@ def init(mode="asyncio", loop=None):
         scheduler.register_asyncio(loop)
 
     elif mode == "rendercanvas":
+        if loop is None:
+            raise TypeError("A loop object is required for the 'rendercanvas' mode")
         scheduler.register_rendercanvas(loop)
 
 
@@ -28,6 +30,7 @@ def loop_factory():
     that observ schedules for async functions and callbacks.
     """
     loop = asyncio.new_event_loop()
-    if hasattr(asyncio, "eager_task_factory"):
-        loop.set_task_factory(asyncio.eager_task_factory)
+    eager_task_factory = getattr(asyncio, "eager_task_factory", None)
+    if eager_task_factory is not None:
+        loop.set_task_factory(eager_task_factory)
     return loop
--- a/observ/proxy.py
+++ b/observ/proxy.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 from copy import copy, deepcopy
-from functools import partial
-from typing import Generic, Literal, TypedDict, TypeVar, cast
+from typing import TYPE_CHECKING, Generic, TypeVar, cast
 
 from .proxy_db import proxy_db
+
+if TYPE_CHECKING:
+    from typing import TypedDict
 
 T = TypeVar("T")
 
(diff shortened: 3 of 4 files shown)


berendkleinhaneveld (approved):

Korijn:
## A mistake I made while preparing this PR, and how it was caught

Leaving this here for the record rather than quietly editing history, since the diff you're reviewing was shaped by a real error and a useful correction.

### What I did wrong

While doing the type-hinting work in #185, I annotated `Watcher._run_callback`, whose traceback-introspection code originally looked like this (this is master, unchanged):

```python
try:
    _run_callback_is_top_frame = (
        e.__traceback__.tb_frame.f_code == self._run_callback.__code__
    )
    _no_lower_frames = e.__traceback__.tb_next is None
    wrong_number_of_arguments = _run_callback_is_top_frame and _no_lower_frames
except AttributeError:
    pass
```

`e.__traceback__` is typed `TracebackType | None`, and a type checker can't tell that the `except AttributeError` branch already handles the `None` case — `try/except` isn't a narrowing construct. So I "cleaned it up" into an explicit check:

```python
tb = e.__traceback__
if tb is not None:
    _run_callback_is_top_frame = tb.tb_frame.f_code == self._run_callback.__code__
    _no_lower_frames = tb.tb_next is None
    wrong_number_of_arguments = _run_callback_is_top_frame and _no_lower_frames
```

This introduced a real bug: `tb.tb_frame` **is the current frame** (the traceback you get from `e.__traceback__` inside the frame that caught the exception always points back at that frame). Binding it to a local (`tb`) creates a frame↔traceback reference cycle, which — since that frame also holds `self` (the `Watcher`) as a local — keeps the watcher alive until the next incidental `gc` pass instead of being reclaimed immediately by refcounting. Two tests that assert immediate weakref-based cleanup after `del watcher` failed.

I "fixed" this by adding `del tb` right after use, and described it in my commit message and PR body as a genuine bug fix — a `functools.partial`-adjacent, pre-existing correctness issue that this PR was cleaning up. **That framing was wrong.** The bug did not exist before my rewrite. The original code never binds `e.__traceback__` to a persistent local — it only reaches it through `e`, and Python automatically deletes `e` at the end of `except ... as e:` for exactly this reason (that's documented CPython behavior, specifically to prevent this class of cycle). My rewrite discarded that free, language-guaranteed protection and replaced it with a hand-maintained `del` that a future maintainer could easily forget, misplace, or bypass (e.g. by adding an early return, or moving code around) without any test failing until someone happened to check object lifetimes without calling `gc.collect()` first.

### How it was caught

Two rounds of pushback from the user:

1. **"I don't see how deleting the traceback object changes anything... wouldn't the local just go out of scope naturally?"** — a fair challenge to verify the mechanism was even real, rather than accept my explanation at face value. I confirmed the cycle with a minimal repro (`gc.disable()`, weakref, check collection before/after `gc.collect()`) — the cycle is real, but this prompted me to look more carefully at *why* the original code didn't have it.
2. **"The new code is more verbose and apparently requires explicit extra code to prevent a ref cycle, while the original did not... it doesn't feel like the higher complexity is justified by typing."** — this is the observation that actually mattered. It was right: the added fragility (a manually-placed `del`, easy to break silently) wasn't a good trade for satisfying a type checker, especially once it became clear the "bug" was self-inflicted rather than pre-existing.

### Where things landed

- **This PR (#184)** no longer touches `_run_callback` at all. It's byte-for-byte identical to master there.
- **#185** keeps the original `try`/`except AttributeError` control flow completely intact, and instead suppresses the two now-flagged lines with narrow `# ty: ignore[unresolved-attribute]` comments plus a short explanatory note. I confirmed the ignores are load-bearing (removing them reproduces the exact two diagnostics) rather than leftover noise.

Net result: no behavior change here, no hand-maintained cycle-breaking anywhere, and the type-checker accommodation in #185 is scoped to exactly the two lines that need it, instead of a control-flow rewrite that traded one footgun for another.

---

Korijn:
I made fable post the above message to expose how ******* stubborn and confident these models are, and how quickly we can be charmed by their "fame" on the internet as the holy grail of all LLMs. Stay sharp people :)

PR #188, opened 2026-07-09
Implement trigger_ref to force-notify watchers of a proxy (closes #123)

### LINKED ISSUE #123: trigger_ref
https://vuejs.org/api/reactivity-advanced.html#triggerref

Korijn:
Closes #123

Modeled after Vue's triggerRef: after making deep mutations to a value
nested inside a shallow proxy (which bypass the proxy and are therefore
invisible to observ), trigger_ref(proxy) notifies the proxy's dep and
all live keydeps, as if its first level was written to. Works on any
proxy; all views of the same target share their deps, so triggering
one view notifies watchers on all of them.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BAdkw8VAYycyoPWjFf2WD7

### DIFF
--- a/tests/test_trigger_ref.py
+++ b/tests/test_trigger_ref.py
@@ -1,0 +1,126 @@
+from unittest.mock import Mock
+
+import pytest
+
+from observ import (
+    computed,
+    readonly,
+    shallow_reactive,
+    trigger_ref,
+    watch,
+    watch_effect,
+)
+
+
+def test_trigger_ref_shallow_key_watcher():
+    state = shallow_reactive({"value": {"count": 0}})
+    watcher = watch(lambda: state["value"], Mock(), sync=True)
+
+    # Deep mutations bypass a shallow proxy, so no watcher fires
+    state["value"]["count"] = 1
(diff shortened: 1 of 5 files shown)


tests/test_trigger_ref.py
```diff
@@ -0,0 +1,126 @@
...
+        trigger_ref({"value": 1})
+
+    with pytest.raises(TypeError):
+        trigger_ref(None)
+
+
+def test_trigger_ref_no_watchers():
+    # Triggering a proxy nobody watches is a no-op
+    state = shallow_reactive({"value": 1})
+    trigger_ref(state)
```
berendkleinhaneveld, line 126: Do we really need this? 😅
Oh well, I guess it tests that it doesn't raise

berendkleinhaneveld (approved):

tests/test_trigger_ref.py
Korijn: 🤷‍♂️

